PR #25, opened 2019-08-29
Milagro to libp2p

Mikerah:
I will be making smaller PRs going forward.

This PR makes the change from milagro-crypto to libp2p-crypto.
A lot of the code doesn't compile yet. This will be addressed in subsequent PRs.
I am also seeking feedback on the design of this as well.


package.json
```diff
@@ -11,10 +11,12 @@
     "test": "test"
   },
   "scripts": {
-    "lint": "aegir lint",
+    "lint": "tslint --fix src/*",
```
mpetrunic: --fix is not ok, because CI might fix something and pass but it will never get commited.

Any reason for switching from aegir?

mpetrunic:
there is a conflict as well

package.json
Mikerah: When I simply had aegir for linting, it didn't lint properly and it showed on github. So, I specified the linter to use in order to get around that.

PR #33, opened 2019-09-20
Initial implementation

wemeetagain:


.travis.yml
```diff
@@ -0,0 +1,51 @@
+language: node_js
+cache: npm
```
mpetrunic: yarn?

mpetrunic: also, to speed up build, specify node_modules directory like in Lodestar so it skips installing if everything was installed in some previous build

package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "libp2p-discv5",
+  "version": "0.0.1",
+  "description": "An implementation of Discovery V5 using Libp2p components",
+  "main": "index.js",
+  "files": [
```
mpetrunic: missing types file location

package.json
```diff
@@ -0,0 +1,78 @@
...
+    "tslint": "^5.18.0",
+    "typescript": "^3.5.3"
+  },
+  "dependencies": {
+    "@chainsafe/milagro-crypto-js": "^0.1.3",
+    "bcrypto": "^4.1.0",
+    "crypto": "^1.0.1",
+    "debug": "^4.1.1",
+    "dgram": "^1.0.1",
+    "es6-promisify": "^6.0.2",
```
mpetrunic: can we not use both es6-promisify and promisify-es6 here as well? :sweat_smile:

package.json
```diff
@@ -0,0 +1,78 @@
...
+    "js-sha256": "^0.9.0",
```
mpetrunic: I'm hearing as-sha256 is pretty much standard nowdays :stuck_out_tongue:

package.json
```diff
@@ -0,0 +1,78 @@
...
+  "author": "ChainSafe Systems",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/ChainSafe/libp2p-discv5/issues"
+  },
+  "homepage": "https://github.com/ChainSafe/libp2p-discv5#readme",
+  "devDependencies": {
+    "@types/chai": "^4.2.0",
+    "@types/mocha": "^5.2.7",
+    "@types/node": "^12.0.10",
```
mpetrunic: @polkadot/ts might help with some of types down there

src/enr/enr.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  /**
+   * constructor
+   */
+  constructor(enrKeyPair?: ENRKeyPair) {
+     this.enrKeyPair = ENRKeyPair;
+     this.keyPairs.set("secp256k1", this.enrKeyPair.compressedPublicKey.toString());
+     this.nodeId = enrKeyPair.derive();
+
+     let peerIdBuf = Buffer.from(sha256(this.keyPairs.get("secp256k1")), "hex");
```
mpetrunic: secp256k1 should be constant

src/enr/enr.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  /**
+   * Returns a RLP encoding of an ENR
+   */
+  public encode(): Buffer {
+    this.content = [
+      RLP.encode(this.sequence),
+      "id", this.keyPairs.get("id"),
+    ];
+    if (this.keyPairs.get("secp256k1")) {
```
mpetrunic: I'm pretty sure those if statemts could be reduce to ["ip", "tcp", ...].foreach

src/enr/enr.ts
```diff
@@ -0,0 +1,235 @@
...
+    enr.signature(RLP.decode(decodedRecord[0]));
+    enr.sequenceNumber(RLP.decode(decodedRecord[1]));
+    enr.id(decodedRecord[2]);
+
+    const indexOfCompPubKey = decodedRecord.indexOf("secp256k1");
+    if (indexOfCompPubKey !== -1) {
+
+    }
+
+    const indexOfIp = decodedRecord.indexOf("ip");
```
mpetrunic: same as for encoding, bunch of duplicated ifs to single foreach

src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+
+import { nextTick } from "async/nextTick";
+import { debug } from "debug";
+import { dgram } from "dgram";
+import { EventEmitter } from "events";
+import { KadDHT } from "libp2p-kad-dht";
+import { multiaddr } from "multiaddr";
+import { PeerId } from "peer-id";
+import { PeerInfo } from "peer-info";
+const log = debug("discv5");
```
mpetrunic: something to consider, all libp2p modules have libp2p: prefix so you can enable all logs with libp2p:*

src/packet/types.ts
```diff
@@ -0,0 +1,67 @@
+// DISCV5 message packet types
+
+import { EthereumNodeRecord } from "../enr/enr";
+
+export type Tag = Buffer; // TAG_LENGTH
+export type Nonce = Buffer; // ID_NONCE_LENGTH
+export type AuthTag = Buffer; // AUTH_TAG_LENGTH
+export type Magic = Buffer; // MAGIC_LENGTH
```
mpetrunic: I would consider moving all types to single file somewhere so it's easier to navigate

package.json
Mikerah: Is @polkadot/ts up to date?

package.json
Mikerah: My bad. I keep misrembering the order of the words es6 and promisify.

src/index.ts
Mikerah: I think we should simply integrate the winston logger here and remove the dependency on the logger currently in use.

package.json
mpetrunic: Not sure but libp2p usually tries to be backwards compatible so api should be pretty much the same. I might be wrong, but ci might complain since strict is enabled in tsconfig

src/index.ts
mpetrunic: Usually libraries tend to use debug. Would be nice for libp2p-discv5 to comply with rest of libp2p libraries, don't know. @wemeetagain ?

mpetrunic:
@wemeetagain ? :joy:

wemeetagain:
idk how it got closed, but I'm intending to open a series of smaller PRs rather than try one big PR that does everything

PR #34, opened 2019-10-06
Refactor

Mikerah:
I made some changes in order to make this codebase more manageable. I refactored most of the current code. I think that the session code that is currently here will need to be completely overhauled as it is poorly written and unmanageable. Additionally, I converted tslint to eslint @mpetrunic's suggestion and have converted from npm to yarn for ease of use. I didn't change the travis file so the build will fail as the file doesn't take into the account the changes made.


GregTheGreek:
Can we close this? @wemeetagain

PR #35, opened 2019-11-14
Add enr implementation

wemeetagain:
Basic ENR implementation
follows [EIP778](https://eips.ethereum.org/EIPS/eip-778)
uses included test vector


src/enr/crypto.ts
```diff
@@ -0,0 +1,63 @@
+// @ts-ignore
+import keccak from "keccakjs";
```
mpetrunic: It would probably be nicer to declare module type like
```javascript
//src/types/keccakjs/index.d.ts
declare module "keccakjs" {
   export default class Keccak...
}
```

src/enr/enr.ts
```diff
@@ -0,0 +1,48 @@
+import assert = require("assert");
```
mpetrunic: from instead of require

mpetrunic: I don't think you have assert declared as dependency (it works in tests because it uses mocha magic assert)

src/enr/crypto.ts
```diff
@@ -0,0 +1,63 @@
...
+import randomBytes from "randombytes";
+import * as secp256k1 from "secp256k1";
```
mpetrunic: both keccak and secp256k1 are already in bcrypto.

Maybe we don't need all those crypto libraries, not sure where else bcrypto is used

wemeetagain: also apparently random bytes :joy: now we just need types -- will investigate!

src/enr/crypto.ts
```diff
@@ -0,0 +1,55 @@
+import keccak = require("bcrypto/lib/keccak");
+import secp256k1 = require("bcrypto/lib/secp256k1");
```
mpetrunic: why not enable esModuleInterop since it's default babel behavior? :)

src/enr/crypto.ts
```diff
@@ -0,0 +1,55 @@
...
+export class ENRKeyPair {
+  public readonly nodeId: NodeId;
+  public readonly privateKey: PrivateKey;
+  public readonly publicKey: PublicKey;
+
+  public constructor() {
+    this.privateKey = createPrivateKey();
+    this.publicKey = publicKey(this.privateKey);
+    this.nodeId = nodeId(this.publicKey);
+  }
```
mpetrunic: Feels like you could use constructor which accepts private key (just a guess)

wemeetagain: Ah, yeah, will add. Currently not using this class, but it could be useful.

src/enr/crypto.ts
wemeetagain: I was thinking of just starting with typescript here, and trying w default settings, making more complicated only as necessary.

src/enr/types.ts
```diff
@@ -0,0 +1,11 @@
+// Custom and aliased types for ENRs
+
+export type NodeId = Buffer;
+export type PrivateKey = Buffer;
+export type PublicKey = Buffer;
+export type Signature = Buffer;
+export type SequenceNumber = bigint;
```
ec2: love that bigint

test/enr.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it("should encode to RLP encoding", () => {
+    const decoded = decode(encode(record, privateKey, seq));
+    for (const [k, v] of decoded.entries()) {
+      expect(v).to.deep.equal(record.get(k));
+    }
+  });
+
+  it("should encode to text encoding", () => {
```
ec2: missing tests?

test/enr.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it("should decode from RLP encoding", () => {
```
ec2: missing tests?

test/enr.test.ts
wemeetagain: I should remove those empty tests.
We'll be testing a roundtrip encode/decode on text and RLP encoding

wemeetagain: Thanks, turns out we weren't testing text encoding, and there's a fun little inconsistency we were hitting: https://github.com/nodejs/node/issues/26512

mpetrunic:
@wemeetagain CI is failing, some type error. I'll try to retrigger CLA

PR #36, opened 2019-11-14
Add packet encoding/decoding

wemeetagain:
Add packet encoding and decoding according to the [discv5 wire protocol](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md)
Test using included [test vectors](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire-test-vectors.md)

PR #37, opened 2019-11-20
Add udp transport

wemeetagain:
Depends on #36 
Add a UDP transport that receives/decodes UDP packets into (sender, `PacketType`, `Packet`)
and encodes/sends (recipient, `PacketType`, `Packet`) via UDP.

### DIFF
--- a/src/packet/decode.ts
+++ b/src/packet/decode.ts
@@ -23,6 +23,8 @@ import {
 /**
  * Decode raw bytes into a packet. The `magic` value (SHA2256(node-id, b"WHOAREYOU")) is passed as a parameter to check
  * for the magic byte sequence.
+ *
+ * Note: this function will modify the input data
  */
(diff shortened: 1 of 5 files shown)


src/packet/decode.ts
```diff
@@ -0,0 +1,111 @@
...
+export function decode(data: Buffer, magic: Magic): [PacketType, Packet] {
+  if (data.length > MAX_PACKET_SIZE) {
+    throw new Error(ERR_TOO_LARGE);
+  }
+  // ensure the packet is large enough to contain the correct headers
+  if (data.length < TAG_LENGTH + AUTH_TAG_LENGTH + 1) {
+    throw new Error(ERR_TOO_SMALL);
+  }
+  const tag = data.slice(0, TAG_LENGTH);
+  data = data.slice(TAG_LENGTH);
```
mpetrunic: I would leave comment on method that data input will be modified or create new buffer from input(performance issue maybe)

PR #38, opened 2019-12-03
Refactor ENR as a class

wemeetagain:
`ENR` is now a class the extends `Map`.
This allows us to maintain a sequence number that gets incremented on every `set`.
It also simplifies/standardizes accessing the `record.id`, `record.publicKey`, and `record.seq`.

The aliases `PublicKey`, `PrivateKey`, and `Signature` are also removed in favor of using `Buffer` directly.

### DIFF
--- a/src/enr/enr.ts
+++ b/src/enr/enr.ts
@@ -5,64 +5,84 @@ import * as RLP from "rlp";
 
 import { MAX_RECORD_SIZE } from "./constants";
 import * as v4 from "./v4";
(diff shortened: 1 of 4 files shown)


mpetrunic:
@wemeetagain whats the reasoning for removing those aliases?

wemeetagain:
I wasn't using them consistently, and it gets a little confusing because we end up dealing with multiple types of public and private key types in different contexts (of differing lengths, etc).

PR #39, opened 2019-12-10
Add generic Keypair interface

wemeetagain:
`KeypairType` will be used to switch between the different types of (eventually) supported keypairs.
Will be used in a PR coming soon.


src/keypair/crypto.ts
```diff
@@ -0,0 +1,71 @@
...
+  return {
+    type,
+    privateKey,
+    publicKey,
+  };
+}
+
+export function privateKeyVerify(keypair: IKeypairPrivate): boolean {
+  switch (keypair.type) {
+    case KeypairType.secp256k1:
```
mpetrunic: I would maybe think about some classes to avoid duplicating that switch case. If you wan't to add another keytype you have to add case clause in like 6-7 places

wemeetagain: :+1:  take another look :)

PR #40, opened 2019-12-10
Add more packet functions

wemeetagain:
Add functions to create packets, encode/decode auth response

### DIFF
--- a/src/packet/create.ts
+++ b/src/packet/create.ts
@@ -1,0 +1,58 @@
+import { randomBytes } from "bcrypto/lib/random";
+import sha256 = require("bcrypto/lib/sha256");
+
+import { AUTH_TAG_LENGTH, WHOAREYOU_STRING, ID_NONCE_LENGTH } from "./constants";
+import { IMessagePacket, Tag, AuthTag, IWhoAreYouPacket, IAuthResponse, Nonce, IAuthHeader } from "./types";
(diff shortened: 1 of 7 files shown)


src/packet/create.ts
```diff
@@ -0,0 +1,58 @@
+import { randomBytes } from "bcrypto/lib/random";
+import sha256 = require("bcrypto/lib/sha256");
```
mpetrunic: require => import

src/packet/create.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { AUTH_TAG_LENGTH, WHOAREYOU_STRING, ID_NONCE_LENGTH } from "./constants";
+import { IMessagePacket, Tag, AuthTag, IWhoAreYouPacket, IAuthResponse, Nonce, IAuthHeader } from "./types";
+import { NodeId, SequenceNumber, ENR } from "../enr";
+
+export function createRandomPacket(tag: Tag): IMessagePacket {
+  return {
+    tag,
+    authTag: randomBytes(AUTH_TAG_LENGTH),
+    message: randomBytes(44),
```
mpetrunic: extract 44 to constant as well?

src/packet/create.ts
wemeetagain: I want to experiment with just using typescript for transpilation in this repo, instead of using babel.

PR #42, opened 2019-12-10
Add message encoding/decoding

wemeetagain:


package.json
```diff
@@ -51,11 +51,13 @@
     "typescript": "^3.7.2"
   },
   "dependencies": {
+    "@types/ip": "^1.1.0",
     "assert": "^2.0.0",
     "base64url": "^3.0.1",
     "bcrypto": "^4.2.8",
     "bigint-buffer": "^1.1.5",
     "dgram": "^1.0.1",
+    "ip": "^1.1.5",
```
mpetrunic: pretty sure this doesn't work in browser, as I remember this uses some nodejs native api ("os" maybe)

package.json
```diff
@@ -51,11 +51,13 @@
     "typescript": "^3.7.2"
   },
   "dependencies": {
+    "@types/ip": "^1.1.0",
```
mpetrunic: devDependency?

src/message/types.ts
```diff
@@ -0,0 +1,74 @@
+import { SequenceNumber, ENR } from "../enr";
+
+export type RequestId = bigint;
```
mpetrunic: why do you treat id as number isn't that always 32 random bytes?

Maybe I'm missing something, but it seems like silly reason to drop safari and node 8 as there is no need for bigint here?!

wemeetagain: The spec doesn't define the length :crying_cat_face: 
We're using bigints elsewhere already in this spec (ENR sequence numbers).

It makes sense to use bigints here because the main things we do with requestIds are: storing request metadata keyed by requestId and doing comparisons with other requestIds, both much more suitable for bigint.

mpetrunic: I would still argue that hex encoded byte string would still be better than dropping node 8 and safari :thinking:

wemeetagain: Discv5 is a udp protocol that by design does not support the browser in its current state.
Node 8 reaches the end of maintenance on Dec. 31 2019.
Eth2 is a blockchain protocol that won't ramp up to full production for ~1 year.
Given those constraints, imo, it makes sense to use future-leaning features including bigint.

PR #44, opened 2020-03-11
Add Kademlia and a basic service implementation

atoulme:
This PR adds more to the WIP of this repo by adding a Kademlia routing table and a service implementation that is meant to be leveraged by end users.


mpetrunic:
Lets wait for @wemeetagain to check logic

wemeetagain:
@mpetrunic a quarter of this PR is my WIP session, I think those `import x = require` statements are mine XD, I'd prefer to keep it vanilla for now.
I think we have defined all bcrypto types we're using, as well.

PR #47, opened 2020-03-20
discv5 service

wemeetagain:
open PR for visibility


package.json
```diff
@@ -32,37 +32,34 @@
...
-    "assert": "^2.0.0",
     "base64url": "^3.0.1",
     "bcrypto": "^4.2.8",
     "bigint-buffer": "^1.1.5",
     "debug": "^4.1.1",
     "dgram": "^1.0.1",
-    "ip": "^1.1.5",
+    "is-ip": "^3.1.0",
+    "libp2p-crypto": "^0.17.5",
+    "libp2p-crypto-secp256k1": "^0.4.3",
```
mpetrunic, line 59: I think libp2p-crypto already contains `libp2p-crypto-secp256k1`.
You can use it like `import {keys} from "libp2p-crypto"`

wemeetagain: Thanks. See https://github.com/libp2p/js-libp2p-crypto/pull/173 :joy_cat:

PR #51, opened 2020-04-23
chore: use new peer-discovery interface

vasco-santos:
In the context of deprecating `peer-info` as described on [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589), this PR changes the `peer-discovery` interface to emit an object containing an `id` property and a `multiaddrs` property.

Note: This should be released with a beta tag until we have everything released, otherwise when people try to install this with `js-libp2p` they will have this as the latest code, which will not work until `0.28`.

BREAKING CHANGE: peer event emits an object with id and multiaddr instead of a peer-info

cc @wemeetagain


vasco-santos:
In another context, I will open an issue in this repo to have the tests from `libp2p-interfaces` running here, in order to avoid future issues and guarantee full compatibility.

vasco-santos:
Yes, sorry about that. I did not realize that you were using `yarn-lock`.

Meanwhile, I added the `package-lock` to git ignore and updated the `yarn-lock` file :)

wemeetagain:
v0.2.0 released under beta tag

PR #65, opened 2020-06-28
Add prettier as an eslint plugin

dapplion:
See https://github.com/ChainSafe/lodestar/pull/1008 for the rationale.

**Note**: The `bracketSpacing` style of this repo is different than `ssz` and `lodestar`. I've changed the config to reflect it
https://github.com/ChainSafe/discv5/compare/master...dapplion:dapplion/prettier#diff-d9a52777b34c9360910514f53aad911dR9
Should we keep the same prettier config across all Eth2.0 tooling?

## How to incorporate
I would recommend for a repo maintainer to run
```
./node_modules/.bin/prettier --write '**/*.{ts,md}'
```
on its own, and commit on top of this PR when the repo is ready for the big resulting diff

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,12 +14,14 @@ module.exports = {
   },
   plugins: [
     "@typescript-eslint",
+    "prettier"
   ],
   extends: [
     "eslint:recommended",
-    "plugin:@typescript-eslint/recommended"
+    "plugin:@typescript-eslint/recommended",
   ],
   rules: {
+    "prettier/prettier": "error",
     //doesnt work, it reports false errors
     "constructor-super": "off",
     "@typescript-eslint/class-name-casing": "error",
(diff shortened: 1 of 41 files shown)


wemeetagain:
Good question re: bracket spacing and lodestar-wide style.
It might make sense to discuss this in a lodestar issue, but I can give some context as to why this is different.
I was attempting to write this library in a 'default typescript'. What this meant to me at the time was trying for a clean-slate approach to code style, identifying any peculiar choices of lodestar, and dumping them in favor of a more standard approach. This ended up translating mostly to having a more default tsconfig, though this peculiarity of bracket spacing also emerged.

As far as lodestar-wide styling, would definitely prefer something as 'standard/default-ish' as possible, though I realize much of it comes down to personal preference and current fads.

dapplion:
Since this library has not the same level of activity as `lodestar` these settings can be merge and it will be easy to change to other settings latter after discussing it.

PR #73, opened 2020-07-18
Check ENR ip address on start

### LINKED ISSUE #72: Handle switching networks upon restart
When we switch to a different network and we restart, we cannot complete the handshake anymore if we had an obsolete `multiaddrUDP` of the previous network with the flow:
+  We send a random package to a node
+ Receive WHOAREYOU request from that node
+ We send authentication package to that node along with our ENR
+ That node found that the ENR contains a different MultiAddrUDP than its `lastSeenMultiaddr` and doesn't establish a session
+ Handshake stops from there

So when we start our node, we need to check if our `multiaddrUDP` is valid anymore. If not, set `multiaddrUDP` to undefined and start the handshake like a fresh ENR.

tuyennhv:
resolves #72 

When we switch network, we can't complete the handshake as I found it in `onAuthMessage`, since the `lastSeenMultiaddr` is different to what's inside `this.remoteEnr`

Tested with an incorrect-ip enr.json in lodestar and we can still do the handshake with this fix
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "libp2p-crypto": "^0.17.5",
     "multiaddr": "^7.4.2",
     "peer-id": "^0.13.11",
+    "public-ip": "^4.0.2",
     "rlp": "^2.2.4",
     "strict-event-emitter-types": "^2.0.0"
   }
--- a/src/enr/enr.ts
+++ b/src/enr/enr.ts
@@ -124,10 +124,10 @@ export class ENR extends Map<ENRKey, ENRValue> {
   }
   set multiaddrUDP(multiaddr: Multiaddr | undefined) {
     if (!multiaddr) {
-      this.set("ip", (undefined as unknown) as Buffer);
-      this.set("udp", (undefined as unknown) as Buffer);
-      this.set("ip6", (undefined as unknown) as Buffer);
-      this.set("udp6", (undefined as unknown) as Buffer);
+      this.delete("ip");
+      this.delete("udp");
+      this.delete("ip6");
+      this.delete("udp6");
       return;
     }
     const protoNames = multiaddr.protoNames();
@@ -168,6 +168,10 @@ export class ENR extends Map<ENRKey, ENRValue> {
   }
   set multiaddrTCP(multiaddr: Multiaddr | undefined) {
     if (!multiaddr) {
+      this.delete("ip");
+      this.delete("tcp");
+      this.delete("ip6");
+      this.delete("tcp6");
       return;
     }
     const protoNames = multiaddr.protoNames();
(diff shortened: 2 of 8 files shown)


src/enr/enr.ts
```diff
@@ -124,6 +124,10 @@ export class ENR extends Map<ENRKey, ENRValue> {
   }
   set multiaddrUDP(multiaddr: Multiaddr | undefined) {
     if (!multiaddr) {
+      this.set("ip", (undefined as unknown) as Buffer);
+      this.set("udp", (undefined as unknown) as Buffer);
+      this.set("ip6", (undefined as unknown) as Buffer);
+      this.set("udp6", (undefined as unknown) as Buffer);
```
dapplion, line 130: Is there a type-safe way of achieving the same result? Maybe by setting the IP to an empty buffer and checking by length?

tuyennhv: yeah setting to empty buffer should be good for type safety, @wemeetagain please confirm ..

wemeetagain: we can do `this.delete(key)`

package.json
```diff
@@ -62,6 +62,7 @@
     "libp2p-crypto": "^0.17.5",
     "multiaddr": "^7.4.2",
     "peer-id": "^0.13.11",
+    "public-ip": "^4.0.2",
```
wemeetagain, line 65: I've got some heartburn about using this dependency to determine our IP.
I think we should instead just add some config to turn off the current IP voting mechanism. Then we might better control it on the lodestar level as needed.

We might also look at what lighthouse is doing here. This IP voting behavior was lifted from their discv5 implementation.

tuyennhv: yeah determining the public ip thru a third party like this is not perfect, I used it just because there is no api to know the public ip.
> I think we should instead just add some config to turn off the current IP voting mechanism. Then we might better control it on the lodestar level as needed.

currently we create Discv5 instance through libp2p so I can't think of how to pass additional config to it

> We might also look at what lighthouse is doing here. This IP voting behavior was lifted from their discv5 implementation.

it's good to know. I checked this in the spec but seems like it's not part of it.. Maybe we can look at prysm as well...

tuyennhv: So what I see is prysm always get external ip address at startup and set to its local enr https://github.com/prysmaticlabs/prysm/blob/master/beacon-chain/p2p/service.go#L182 so it'll never have this isslue

tuyennhv: @wemeetagain can you double check with lighthouse, I'm not familiar with it :D

tuyennhv: one possible solution would be to check in PONG message, if the voted ip address is part of network interfaces, we'll continue. Otherwise just ignore.

It means that we'll only insert multiaddrs to ENR if it's a public server (where we have a static public/external ip address). For pc/laptop environment connected to LAN, we'll not insert it.

tuyennhv: had a discussion with @wemeetagain . The final conclusion is to check if public ip address in local network interface, if not just wipe it at lodestar bootup time, not at discv5.

PR #77, opened 2020-08-03
Set ip address to enr at startup

tuyennhv:
resolves https://github.com/ChainSafe/lodestar/issues/1295

It's most likely we'll either configure `bootEnr` as `0.0.0.0` or a public/static ip address instead of DNS or local ip address, so just set to ENR if we found it's not 0.0.0.0


wemeetagain:
Closing in favor of configuration using ENR directly

PR #88, opened 2020-08-14
Add lookupTimeout configuration

wemeetagain:
Add a `lookupTimeout` configuration option.
Sets the maximum amount of time to spend on each lookup. Default to 60 seconds.

### DIFF
--- a/src/kademlia/lookup.ts
+++ b/src/kademlia/lookup.ts
@@ -51,6 +51,9 @@ export class Lookup extends (EventEmitter as { new (): LookupEventEmitter }) {
    */
   config: ILookupConfig;
 
+  timeout?: number;
+  timeoutFn: Function;
+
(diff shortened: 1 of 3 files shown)


src/kademlia/lookup.ts
```diff
@@ -77,6 +82,7 @@ export class Lookup extends (EventEmitter as { new (): LookupEventEmitter }) {
     if (this.state === LookupState.Finished) {
       return;
     }
+    clearTimeout(this.timeout);
```
dapplion, line 85: strictNullChecks police over here. Should be `if (this.timeout) then`

wemeetagain: We have strict null checks turned on, this isn't an issue.

clearTimeout can be passed undefined / invalid timerId w/o issue, part of the spec.

PR #101, opened 2020-09-26
Fix multiaddr port after decoding enr from a txt

tuyennhv:
js-multiaddr 8.0.0 only expects a Buffer with 0 byte offset, this caused discv5 to use incorrect udp port after decoding from enr and cannot do handshake with other nodes. The issue is (for later reference):
+ Discv5 gets inccorrect udp port after decoding an enr from txt
+ It sends random packets to nodes
+ Receive no packet from then

We expect it to be fixed in https://github.com/multiformats/js-multiaddr/pull/146/files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,5 +69,8 @@ lib
 # swp file
 *.swp
 
+# ide
+.vscode
+
 # others
 package-lock.json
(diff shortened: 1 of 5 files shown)


wemeetagain:
Wow :clap: great work. Very subtle bug.

We can merge this now so we can cut a new release. And once multiaddr is re-released, revert this.

wemeetagain (approved):

tuyennhv:
yeah. At that time t I'd like to keep the tests in this PR to make sure we don't face it again

PR #107, opened 2020-11-19
Initial discv5.1 update

wemeetagain:
Tested with discv5-cli against the pyrmont testnet, seems to find peers and interop correctly :)

TODO (not in this PR) - handle newly added TALKREQ / TALKRESP messages - they don't effect discovery, but are an extension that allows discv5 to negotiate messages for some additional application-specific protocol


mpetrunic (approved):
LGTM, but I'm not really familiar discv5 spec

PR #115, opened 2021-04-28
chore: update dependencies

vasco-santos:
Updates dependencies to multiaddr, so that I can update libp2p in lodestar properly and test the new libp2p release

BREAKING CHANGE: new multiaddr used with different API


src/enr/enr.ts
```diff
@@ -1,4 +1,4 @@
-import Multiaddr = require("multiaddr");
+import { Multiaddr, protocols } from "multiaddr";
```
dapplion, line 1: :tada:

src/enr/enr.ts
```diff
@@ -182,7 +182,7 @@ export class ENR extends Map<ENRKey, ENRValue> {
   get tcp6(): number | undefined {
     const raw = this.get("tcp6");
     if (raw) {
-      return muConvert.toString(Multiaddr.protocols.names.tcp.code, raw) as number;
+      return muConvert.toString(protocols.names.tcp.code, raw) as number;
```
dapplion, line 185: Is the casting to number still necessary?

src/service/service.ts
```diff
@@ -1,6 +1,6 @@
 import { EventEmitter } from "events";
 import debug from "debug";
-import Multiaddr = require("multiaddr");
+import { Multiaddr } from "multiaddr";
 import isIp = require("is-ip");
 import PeerId = require("peer-id");
```
dapplion, line 5: PeerId can be imported with `import` already right?

src/enr/enr.ts
vasco-santos: There is a test failing that I could not understand why last night. I think it is related to this now that you point it out. It returns a [String](https://github.com/multiformats/js-multiaddr/blob/master/src/convert.js#L34). But there is these ]local definitions](https://github.com/ChainSafe/discv5/blob/9a336ebfe44068a11b1f54628fc9fdb74ee168d4/types/multiaddr/index.d.ts#L8) which also return a number

src/service/service.ts
vasco-santos: done, also updated in other places

src/enr/enr.ts
vasco-santos: fixed the local type definitions. we do not support types in multiaddr with an import path, as we do in libp2p interfaces. This is because we would not expect these utils to be used externally. Perhaps worth creating a follow up discussion on how we can enable this better, with a multiaddr utils, or exporting in the root

vasco-santos:
@dapplion travis CI is not working, I tried to replace it with github actions like other repos you have. But it is not starting too, can you see if it needs to be enabled?

dapplion:
> @dapplion travis CI is not working, I tried to replace it with github actions like other repos you have. But it is not starting too, can you see if it needs to be enabled?

@wemeetagain can you take a look?

vasco-santos:
Thanks, I ended up rebasing from master which now has actions. We now only support 14+ on libp2p so I updated the actions and also added other OS. This should now be ready for review

dapplion (approved):

wemeetagain (approved):
Thanks @vasco-santos :heart:

PR #119, opened 2021-07-06
fix: decodeFindNode

samlior:
when distance is `256`, the rlp encode result will be `<Buffer 01 00>`. so we need to use `readUIntBE` instead of `readUInt8`.

### DIFF
--- a/src/message/decode.ts
+++ b/src/message/decode.ts
@@ -87,7 +87,7 @@ function decodeFindNode(data: Buffer): IFindNodeMessage {
   if (!Array.isArray(rlpRaw[1])) {
     throw new Error(ERR_INVALID_MESSAGE);
   }
-  const distances = ((rlpRaw[1] as unknown) as Buffer[]).map((x) => (x.length ? x.readUInt8(0) : 0));
+  const distances = ((rlpRaw[1] as unknown) as Buffer[]).map((x) => (x.length ? x.readUIntBE(0, x.length) : 0));
   return {
     type: MessageType.FINDNODE,
     id: toBigIntBE(rlpRaw[0]),


wemeetagain:
I think max distance should be 255? https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#node-table
If thats the case, we should have better validation and throw an Error if the distance is not valid (> 1 byte).

samlior:
The max `distance` is indeed `255`, but the `distance` range in `FindNode` message should be `[0, 256]`.
Because when distance 0 is requested, the result set should contain the recipient's current record.
https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#findnode-request-0x03

```ts
// src/service/service.ts
private onFindNode(srcId: NodeId, src: Multiaddr, message: IFindNodeMessage): void {
    const { id, distances } = message;
    let nodes: ENR[] = [];
    distances.forEach((distance) => {
      // if the distance is 0, send our local ENR
      if (distance === 0) {
        this.enr.encodeToValues(this.keypair.privateKey);
        nodes.push(this.enr);
      } else {
        nodes.push(...this.kbuckets.valuesOfDistance(distance));
      }
    });
    // ...
}
```

```ts
// src/kademlia/kademlia.ts
valuesOfDistance(value: number): ENR[] {
    const bucket = this.buckets[value - 1];
    return bucket === undefined ? [] : bucket.values();
}
```
The index of bucket is `value - 1` instead of `value`, and the length of `this.buckets` is 256, so the range of `value` should be `(0, 256]`. And when `distance` is 0, the node will return his local ENR. 
Therefore, the range of the `distance` in `FindNode` message is `[0, 256]`.

wemeetagain:
Thanks for making that clear for me. :+1:

wemeetagain (approved):

PR #122, opened 2021-07-22
Add handlers/helpers for TALKREQ/TALKRESP message types

acolytec3:
Simple TALKREQ/TALKRESP message handling

- discv5.service emits event when a TALKREQ/TALKRESP message is received 
- `discv5.sendTalkReq` broadcasts a a provided `payload` to all nodes in routing table with a given `protocol` in the `TALKREQ` message format
- `discv5.sendTalkResp` sends a TALKRESP message to a specified `nodeId` with a provided `payload`

### DIFF
--- a/src/message/create.ts
+++ b/src/message/create.ts
@@ -1,7 +1,16 @@
 import { randomBytes } from "bcrypto/lib/random";
 import { toBigIntBE } from "bigint-buffer";
 
-import { RequestId, IPingMessage, MessageType, IPongMessage, IFindNodeMessage, INodesMessage } from "./types";
+import {
+  RequestId,
+  IPingMessage,
+  MessageType,
+  IPongMessage,
+  IFindNodeMessage,
+  INodesMessage,
+  ITalkReqMessage,
+  ITalkRespMessage,
+} from "./types";
 import { SequenceNumber, ENR } from "../enr";
 
 export function createRequestId(): RequestId {
@@ -45,5 +54,21 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
     id,
     total,
     enrs,
+  };
+}
+
+export function createTalkRequestMessage(request: string): ITalkReqMessage {
+  return {
+    type: MessageType.TALKREQ,
+    id: createRequestId(),
+    protocol: Buffer.from("0x01"),
+    request: Buffer.from(request),
+  };
+}
+export function createTalkResponseMessage(message: ITalkReqMessage): ITalkRespMessage {
+  return {
+    type: MessageType.TALKRESP,
+    id: message.id,
+    response: Buffer.from(message.request.toString("utf-8") + " - back to you"),
   };
 }
--- a/src/message/create.ts
+++ b/src/message/create.ts
@@ -57,18 +57,18 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
   };
 }
 
-export function createTalkRequestMessage(request: string | Buffer, protocol: string): ITalkReqMessage {
+export function createTalkRequestMessage(request: string | Buffer, protocol: string | Uint8Array): ITalkReqMessage {
   return {
     type: MessageType.TALKREQ,
     id: createRequestId(),
-    protocol: Buffer.from(protocol.toString()),
+    protocol: Buffer.from(protocol),
     request: Buffer.from(request),
   };
 }
-export function createTalkResponseMessage(request: ITalkReqMessage, payload: Buffer): ITalkRespMessage {
+export function createTalkResponseMessage(requestId: RequestId, payload: Uint8Array): ITalkRespMessage {
   return {
     type: MessageType.TALKRESP,
-    id: request.id,
-    response: payload,
+    id: requestId,
+    response: Buffer.from(payload),
   };
 }
--- a/src/message/create.ts
+++ b/src/message/create.ts
@@ -57,18 +57,18 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
   };
 }
 
-export function createTalkRequestMessage(request: string): ITalkReqMessage {
+export function createTalkRequestMessage(request: string | Buffer, protocol: string): ITalkReqMessage {
   return {
     type: MessageType.TALKREQ,
     id: createRequestId(),
-    protocol: Buffer.from("0x01"),
+    protocol: Buffer.from(protocol.toString()),
     request: Buffer.from(request),
   };
 }
-export function createTalkResponseMessage(message: ITalkReqMessage): ITalkRespMessage {
+export function createTalkResponseMessage(request: ITalkReqMessage, payload: Buffer): ITalkRespMessage {
   return {
     type: MessageType.TALKRESP,
-    id: message.id,
-    response: Buffer.from(message.request.toString("utf-8") + " - back to you"),
+    id: request.id,
+    response: payload,
   };
 }
--- a/src/message/create.ts
+++ b/src/message/create.ts
@@ -57,7 +57,7 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
   };
 }
 
-export function createTalkRequestMessage(request: string | Buffer, protocol: string | Uint8Array): ITalkReqMessage {
+export function createTalkRequestMessage(request: string | Uint8Array, protocol: string | Uint8Array): ITalkReqMessage {
   return {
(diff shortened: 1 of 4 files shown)


acolytec3:
@wemeetagain Here's a first draft of the `TALKREQ`/TALKRESP` message handling.  See [here](https://github.com/acolytec3/ultralight/commit/ee7ea0715722b56a62d678f35c710acfa02c32fb) for an example of usage by a client app.

src/service/service.ts
```diff
@@ -283,6 +287,37 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
     });
   }
 
+  /**
+   * Broadcast TALKREQ message to all nodes in routing table
+   */
+  public async sendTalkReq(payload: Buffer, protocol: string): Promise<void> {
+    const msg = createTalkRequestMessage(payload, protocol);
+    for (const node of this.kadValues()) {
+      console.log(node.nodeId);
```
wemeetagain, line 296: stray console.log

src/message/create.ts
```diff
@@ -47,3 +56,19 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
...
+
+export function createTalkRequestMessage(request: string | Buffer, protocol: string): ITalkReqMessage {
+  return {
+    type: MessageType.TALKREQ,
+    id: createRequestId(),
+    protocol: Buffer.from(protocol.toString()),
+    request: Buffer.from(request),
+  };
+}
+export function createTalkResponseMessage(request: ITalkReqMessage, payload: Buffer): ITalkRespMessage {
```
wemeetagain, line 68: Can we just pass in the request `id` rather than the whole request?

src/service/service.ts
```diff
@@ -283,6 +287,37 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
...
+      } else {
+        log(`Sent TALKREQ message to node ${node.nodeId}`);
+      }
+    }
+  }
+
+  /**
+   * Broadcast TALKRESP message to requestingi node
+   */
+  public async sendTalkResp(srcId: NodeId, request: ITalkReqMessage, payload: Buffer): Promise<void> {
```
wemeetagain, line 309: We can just pass in the request id here, too.

src/service/types.ts
```diff
@@ -20,6 +20,18 @@ export interface IDiscv5Events {
    * Our local ENR IP address has been updated
    */
   multiaddrUpdated: (addr: Multiaddr) => void;
+  /**
+   * A TALKREQ message was received. Messages
```
wemeetagain, line 24: stray word

src/service/service.ts
```diff
@@ -283,6 +287,37 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
     });
   }
 
+  /**
+   * Broadcast TALKREQ message to all nodes in routing table
+   */
+  public async sendTalkReq(payload: Buffer, protocol: string): Promise<void> {
```
wemeetagain, line 293: maybe we can change the name of this method to something that more clearly evokes 'sending messages to all nodes'.
Eg: `broadcastTalkReq`.

I'd imagine that we would eventually also have a `sendTalkReq(srcId: NodeId, protocol: string, payload: Buffer): Promise<void>` that would mirror `sendTalkResp` for sending a TALKREQ to a single peer.

src/service/service.ts
```diff
@@ -283,6 +287,37 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
...
+   */
+  public async sendTalkResp(srcId: NodeId, request: ITalkReqMessage, payload: Buffer): Promise<void> {
+    const msg = createTalkResponseMessage(request, payload);
+    const enr = this.getKadValue(srcId);
+    const addr = await enr?.getFullMultiaddr("udp");
+    if (enr && addr) {
+      await this.sessionService.sendResponse(addr, srcId, msg);
+      log(`Sent TALKRESP message to node ${enr.id}`);
+    } else {
+      log(`Node ${srcId} not found`);
```
wemeetagain, line 317: I wonder if its worth splitting out 'node not found' from 'no ip + udp port found'

src/service/service.ts
```diff
@@ -283,6 +287,37 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
...
+      if (!sendStatus) {
+        log(`Failed to send TALKREQ message to node ${node.nodeId}`);
+      } else {
+        log(`Sent TALKREQ message to node ${node.nodeId}`);
+      }
+    }
+  }
+
+  /**
+   * Broadcast TALKRESP message to requestingi node
```
wemeetagain, line 307: typo

src/message/create.ts
```diff
@@ -47,3 +56,19 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
     enrs,
   };
 }
+
+export function createTalkRequestMessage(request: string | Buffer, protocol: string): ITalkReqMessage {
```
wemeetagain, line 60: I think protocol should also be allowed to be a Uint8Array? Seems that the 'protocol' can be any bytestring.

src/service/service.ts
wemeetagain: Can we use `Uint8Array` instead of `Buffer` for these added function params? Here and in the other added methods/functions. We can use `Buffer` internally, but it's nice to allow consumers to pass in native `Uint8Array`, rather than need to wrap in a `Buffer`.

src/service/service.ts
```diff
@@ -643,6 +682,16 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
     }
   };
 
+  private onTalkReq = (srcId: NodeId, src: Multiaddr, message: ITalkReqMessage): void => {
+    log("Received TALKREQ message from Node: %s", srcId);
+    this.emit("talkReqReceived", srcId, message);
```
wemeetagain, line 687: I wonder if all available data should be emitted.
eg: `this.emit("talkReqReceived", srcId, src, message);`
Then again, I'm not sure that the `Multiaddr` is very useful / actionable.
It seems more likely that the ENR will be more useful. :shrug: 
I guess this can be revisited later.

wemeetagain (changes requested):
Generally looking really good.

Just had a few small comments & suggestions.

src/service/service.ts
```diff
@@ -283,6 +288,40 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
...
+  public async sendTalkResp(srcId: NodeId, requestId: RequestId, payload: Uint8Array): Promise<void> {
+    const msg = createTalkResponseMessage(requestId, payload);
+    const enr = this.getKadValue(srcId);
+    const addr = await enr?.getFullMultiaddr("udp");
+    if (enr && addr) {
+      await this.sessionService.sendResponse(addr, srcId, msg);
+      log(`Sent TALKRESP message to node ${enr.id}`);
+    } else {
+      if (!addr && enr) {
+        log(`No ip + udp port found for node ${enr.id}`);
```
wemeetagain, line 318: I think we want to log the node `srcId` and not `enr.id`

src/message/create.ts
```diff
@@ -47,3 +56,19 @@ export function createNodesMessage(id: RequestId, total: number, enrs: ENR[]): I
     enrs,
   };
 }
+
+export function createTalkRequestMessage(request: string | Buffer, protocol: string | Uint8Array): ITalkReqMessage {
```
wemeetagain, line 60: Buffer -> Uint8Array

src/service/service.ts
```diff
@@ -283,6 +288,40 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
     });
   }
 
+  /**
+   * Broadcast TALKREQ message to all nodes in routing table
+   */
+  public async broadcastTalkReq(payload: Buffer, protocol: string | Uint8Array): Promise<void> {
```
wemeetagain, line 294: Buffer -> Uint8Array

wemeetagain (changes requested):

src/service/service.ts
acolytec3: Sorry brain fart on that one.  I've corrected both this and the one above and verified that everything is still working with my client app so far.

wemeetagain (approved):
:tada: Thanks @acolytec3 
I'll publish to NPM shortly

PR #128, opened 2021-07-27
Make searchInterval optional, handle infinity case

wemeetagain:
In order to avoid disruption to the exposed interface, make `searchInterval` optional, with the default being the pre-existing behavior.

Also, properly handle `Infinity`, which shouldn't cause an infinite sleep, but rather disable auto lookup entirely.

### DIFF
--- a/src/libp2p/discv5.ts
+++ b/src/libp2p/discv5.ts
@@ -9,6 +9,11 @@ import { createNodeId, ENR } from "../enr";
 import { IDiscv5Config } from "../config";
 import { toBuffer } from "../util";
 
+// Default to 0ms between automatic searches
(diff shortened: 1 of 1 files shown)


src/libp2p/discv5.ts
```diff
@@ -63,7 +67,7 @@ export class Discv5Discovery extends EventEmitter {
       config: options,
       metrics: options.metrics,
     });
-    this.searchInterval = options.searchInterval;
+    this.searchInterval = options.searchInterval ?? 0;
```
dapplion, line 70: Declare at the top of the file as `DEFAULT_SEARCH_INTERVAL_MS` and explain why this number and our roadmap to fine tune it

dapplion (approved):

PR #133, opened 2021-08-31
fix src value in WHOAREYOU error messages

acolytec3:
The `src` in the WHOAREYOU error messages is adjusted be the `toString()` value of the source `multiaddr`.  Otherwise the error messages look like

```js
Received a WHOAREYOU packet that references an unknown or expired request. source: Multiaddr, token: e022fd0369fa83381b0bcf74
```

Fixed version now looks like:
```js
Received a WHOAREYOU packet that references an unknown or expired request. source: /ip4/0.0.0.0/tcp/5500/wss, token: 6c22f64bf71b9d81ff3bfef9
```


wemeetagain (approved):
Thanks

PR #136, opened 2021-09-02
Convert `broadcastTalkReq` to return promise

acolytec3:
The `broadcastTalkReq` message implemented by #122 doesn't return anything at present so the parent application isn't currently able to match up messages sent with `TALKRESP` messages received from the network.

This PR converts `broadcastTalkReq` to a promise based method that returns the first response it receives from the network or else rejects after a 1 second timeout.

### DIFF
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -304,12 +304,23 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
    */
   public async broadcastTalkReq(payload: Buffer, protocol: string | Uint8Array, timeout = 1000): Promise<Buffer> {
     return await new Promise((resolve, reject) => {
+      const listenerId = this.listeners("talkRespReceived").length;
       const msg = createTalkRequestMessage(payload, protocol);
-      const responseTimeout = setTimeout(() => reject("Request timed out"), timeout);
+      const responseTimeout = setTimeout(() => {
+        try {
+          const listener = this.listeners("talkRespReceived")[listenerId];
+          this.removeListener("talkRespReceived", listener as () => void);
+        } catch {
+          // Just catching any error if listener is already removed
+        }
+        reject("Request timed out");
+      }, timeout);
       this.on("talkRespReceived", (srcId, enr, res) => {
+        const listener = this.listeners("talkRespReceived")[listenerId];
         if (res.id === msg.id) {
           clearTimeout(responseTimeout);
           resolve(res.response);
+          this.removeListener("talkRespReceived", listener as () => void);
         }
       });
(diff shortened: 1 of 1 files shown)


acolytec3:
@wemeetagain Any thoughts on this. I'm using it quite successfully in my fork and I believe it addresses some feedback from the original TALKREQ PR around a promise based interface.

src/service/service.ts
```diff
@@ -300,18 +300,28 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
...
-      const sendStatus = this.sendRequest(node.nodeId, msg);
-      if (!sendStatus) {
-        log(`Failed to send TALKREQ message to node ${node.nodeId}`);
-      } else {
-        log(`Sent TALKREQ message to node ${node.nodeId}`);
+  public async broadcastTalkReq(payload: Buffer, protocol: string | Uint8Array, timeout = 1000): Promise<Buffer> {
+    return await new Promise((resolve, reject) => {
+      const msg = createTalkRequestMessage(payload, protocol);
+      const responseTimeout = setTimeout(() => reject("Request timed out"), timeout);
+      this.on("talkRespReceived", (srcId, enr, res) => {
```
wemeetagain, line 309: We need to remove the event listener in either case so they don't build up.

acolytec3: Does it make sense that a node would ever have more than one `TALKREQ` message waiting for a response?  It's not 100% obvious to me that we would ever want to send more than one at a time but if so, that makes clearing the right listener a little trickier.  I was contemplating creating a dynamic message type of `TALKREQ-${messageId}`in this instance as the most straightforward way to track `TALKREQ` messages using the promise based interface vs just broadcasting messages out and listening for general responses.  That would change the API somewhat though.  In this case, I'm thinking to revert my changes on this message and create a new `broadcastTalkReqMessagePromise` or something similar that uses the new dynamically named message type so that I can clear the listeners when the promise resolves/rejects.  Thoughts?

acolytec3:
What do you think about the promise resolving with the full `ITalkRespMessage` instead of just the response?  I'm debating whether it's ever useful to have the request ID in the resolved promise.

wemeetagain:
I think its good as it is without the full Message object. The request id is only useful for connecting the response to the request, and we're bridging that gap for the consumer with these promises.

wemeetagain (approved):

acolytec3:
Sounds good. I'm going to call this done at this point unless there's anything else you need added.

wemeetagain:
Yup lgtm, gonna merge and cut a release :)

PR #142, opened 2021-10-15
Add multiaddr to ENR

D4nte:


D4nte:
This was meant for my own fork

dapplion:
Feel free to do a PR upstream if you feel a feature is missing!

wemeetagain:
> What are your thoughts in regards of the cryptography backend? Bcrypto vs secpk1? Was bcrypto chosen because it supports various curves, even though only secpk1 is used for ENRs?

I think bcrypto was used because it was already being used for discv5 crypto.
My hope was ideally we could use libp2p-crypto.

Issue #145, opened 2021-11-02
NodeId performance


tuyennhv:
**Description**
This shows that lodestar node uses 23% of its time just for getting node id from enr
[1101_pithos_contabo_3.cpuprofile.zip](https://github.com/ChainSafe/discv5/files/7457405/1101_pithos_contabo_3.cpuprofile.zip)
<img width="1115" alt="Screen Shot 2021-11-02 at 09 12 10" src="https://user-images.githubusercontent.com/10568965/139777468-ebeb552a-3aea-44fd-a716-ef3e8854ec61.png">

we should either cache node id in enr or manage enrs better (index it) in bucket

PR #146, opened 2021-11-02
fix: index bucket by node id

tuyennhv:
**Motivation**
+ Issue #145 shows that there's a performance issue searching for enr by node id

**Description**
+ Index bucket by node id, similar to forkchoice
+ When we want to find enr by node id, just find the index `this.bucketIndices` first
+ The benchmark shows that it's 9x better even we have to update the index frequently. Updating the map is still way faster than extracting node id from enr
+ Add unit tests

part of #145

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "build": "tsc --declaration --outDir lib",
     "prepublishOnly": "yarn build",
     "lint": "eslint --color --ext .ts src/",
-    "test": "mocha -r ts-node/register 'test/**/*.test.ts'"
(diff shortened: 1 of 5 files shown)


mpetrunic:
@tuyennhv is this effort abandoned. Do we wanna release this? Conflicts should be fixed

tuyennhv:
> @tuyennhv is this effort abandoned. Do we wanna release this? Conflicts should be fixed

discv5 has a new implementation since then, didn't notice this issue any more so I should close this PR, thanks for reminding me @mpetrunic

PR #151, opened 2021-11-10
AddrVotes: Fix remoteVote callback

tuyennhv:
**Motivation**
+ a regression of #144
+ `removeVote` was called with undefined "this"

**Description**
+ rewrite `removeVote` to fix the undefined "this" issue
+ write unit test to reproduce and fix the issue


tuyennhv:
replaced by #152

PR #155, opened 2021-11-22
Refresh implementation

wemeetagain:
- Add limit to session cache size
- Simplify session logic (only create fully-connected session objects)
- Update kademlia implementation (return values that distinguish between different insert/update outcomes)
- Remove unknown ENR code paths, instead use `NodeContact` which allows connecting to a multiaddr or ENR
- Simplify TALKREQ promise handling, allow for attaching callbacks to any request
- Add e2e tests to test various scenarios (simple local tests, TALKREQ/TALKRESP, find peers from mainnet bootnodes)


wemeetagain:
This branch gets significantly more FINDNODES requests than the lastest release

This branch:
Screenshot from 2021-11-26 09-54-50

Latest release:
Screenshot from 2021-11-26 09-53-06

dapplion (approved):

src/kademlia/kademlia.ts
```diff
@@ -27,70 +26,156 @@ export class KademliaRoutingTable extends (EventEmitter as { new (): BucketEvent
    * @param localId the ID of the local node
    * @param k the size of each bucket (k value)
```
acolytec3, line 27: Did you intend to leave this `k` param definition here since it was removed from the constructor?  Just curious as I was doing some work on my portalnetwork module and noticed this k parameter was no longer passed to the table constructor and it looks like you are no longer limiting the size of each bucket.  Is that accurate?

wemeetagain: It was meant to be hardcoded to a constant, looks like I missed this

acolytec3: Ah, so is discv5 generally moving to a practice of fixed sized buckets? Just curious as it will have downstream effects (probably minimal but still impacts) on Ultralight

wemeetagain: This shouldn't have any downstream effects (unless you need to configure the size of the buckets differently).
But afaik everything uses the same size bucket.

acolytec3: Okay, there haven't been any final decisions about bucket sizes yet for the portal network but will keep this in mind once we get there.

PR #166, opened 2022-03-21
fix: add toNewUint8Array when decoding enr values

samlior:
There is now a problem with encoding and decoding enr addresses because `toNewUint8Array` should be called before `muConvert.toString`


dapplion (approved):

samlior:
The same problem, when decoding the message

wemeetagain (approved):
Thanks @samlior

samlior:
You're welcome @wemeetagain , I just want this project to be better. Would you or your team have time to take a look at #167. Because I found more issues with ipv6 and I don't know how to fix it.

PR #174, opened 2022-05-05
chore: Bump peer-id dep to latest

acolytec3:
Bumping this dependency upgrades several transitive dependencies down the dependency tree (notable `libp2p/crypto` which in turn switches from the deprecated `noble-secp256k1` to the updated/namespaced `@noble/secp256k1`).  The only potentially impactful breaking change here is `peer-id` v0.16.0 requires Node v15+.  We've been using this version of `peer-id` in the `portalnetwork` module for a while now with no ill effects and I'd like to remove the deprecated `noble-secp256k1` module from our dependency tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "scripts": {
     "check-types": "tsc --noEmit",
     "build": "tsc --declaration --outDir lib",
+    "prepare": "npm run build",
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
LGTM

PR #178, opened 2022-05-11
fix: A more robust check for rinfo.family 

### LINKED ISSUE #176: Have a more robust check for rinfo.family
As reported on discord [here](https://discord.com/channels/593655374469660673/743858262864167062/973365618772439070) in node version 18 on MacOS, the value of rinfo.family was an integer type 4.

Even though the documentation [here](https://nodejs.org/api/net.html#net_socketaddress_family) states the value should be of type string with value being either 'ipv4' or 'ipv6'. Clearly this is not the case as I was also able to confirm the issue.

This PR makes the check for `rinfo.family` more robust such that it does not fail in this observed mismatch between documentation and behaviour of node/os version.

dadepo:
Closes #176

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        node: [15, 16]
+        node: [16, 17, 18]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2-beta
--- a/src/transport/udp.ts
+++ b/src/transport/udp.ts
@@ -50,7 +50,9 @@ export class UDPTransportService
   }
 
   public handleIncoming = (data: Buffer, rinfo: IRemoteInfo): void => {
-    const multiaddr = new Multiaddr(`/${rinfo.family === "IPv4" ? "ip4" : "ip6"}/${rinfo.address}/udp/${rinfo.port}`);
+    const multiaddr = new Multiaddr(
+      `/${String(rinfo.family).endsWith("4") ? "ip4" : "ip6"}/${rinfo.address}/udp/${rinfo.port}`
+    );
     try {
       const packet = decodePacket(this.srcId, data);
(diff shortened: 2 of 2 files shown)


dapplion (approved):
I can approve the PR but have no way to verify it's correct since I don't own a MacOS. Could you look into adding macos tests to demostrate correctness? Github actions have runners for it I think

dadepo:
> Could you look into adding macos tests to demostrate correctness? Github actions have runners for it I think

I checked, github actions do support MacOS but unfortunately M1 is still in currently in public beta. [See](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)

But the main issue seems to be mainly about node version 18. I updated the version in another branch that does not have the fix and was able to confirm the tests fails. [See](https://github.com/ChainSafe/discv5/pull/179)

Updated the same version in this PR, and all tests passes. [See](https://github.com/ChainSafe/discv5/pull/178/commits/3f1eb441f8d577181854074aa0f4a888ab279865). 

So PR should be good to merge.

dapplion (approved):

mpetrunic:
> > Could you look into adding macos tests to demostrate correctness? Github actions have runners for it I think
> 
> I checked, github actions do support MacOS but unfortunately M1 is still in currently in public beta. [See](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)
> 
> But the main issue seems to be mainly about node version 18. I updated the version in another branch that does not have the fix and was able to confirm the tests fails. [See](https://github.com/ChainSafe/discv5/pull/179)
> 
> Updated the same version in this PR, and all tests passes. [See](https://github.com/ChainSafe/discv5/pull/178/commits/3f1eb441f8d577181854074aa0f4a888ab279865). 
> 
> So PR should be good to merge. 
> 

I can run it locally on m1 if you are concerned

dadepo:
> I can run it locally on m1 if you are concerned

Thanks! 🙏 But that should not be needed anymore. Plus I do have an M1 and was able to verify the issue/fix.

PR #179, opened 2022-05-12
include version 18 of node. See if this trigger the rinfo.family issue

dadepo:


dadepo:
Used to confirm the test fails without the fix in the [PR](https://github.com/ChainSafe/discv5/pull/178)

PR #180, opened 2022-05-18
feat: all unverified inbound sessions

acolytec3:
Features
- Adds new optional `sessionConfig` parameter to allow inbound sessions with invalid ENRs (i.e. where observed socket address doesn't match socket address reported in ENR - to allow nodes reporting local IP/port combinations to receive `discv5` PONG messages in order to update ENRs).
  - Note: nodes with "unverified sessions" are not added to the `discv5` routing table so the discv5 service will never send outbound messages to them to ensure we aren't propagating DDoS attacks
- Allows `sendTalkResp` to specify a target node via session key (`nodeId + socketAddr`) to allow responding to nodes with unverified sessions (so will not have an ENR in the routing table)
- Makes `service.sendPing` public so it can be triggered by portal network layer to trigger `enrUpdate` logic via PONG messages.

Fixes
- Fixes tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chainsafe/discv5",
-  "version": "0.6.7",
+  "version": "0.7.0",
   "description": "Discovery V5",
   "main": "lib/index.js",
   "files": [
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "lint": "eslint --color --ext .ts src/",
     "test": "yarn test:unit && yarn test:e2e",
     "test:unit": "mocha 'test/unit/**/*.test.ts'",
-    "test:e2e": "mocha 'test/unit/**/*.test.ts'"
+    "test:e2e": "mocha 'test/e2e/**/*.test.ts'"
   },
   "pre-push": [
     "lint"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chainsafe/discv5",
-  "version": "0.7.0",
+  "version": "0.7.1",
   "description": "Discovery V5",
   "main": "lib/index.js",
   "files": [
@@ -72,7 +72,7 @@
     "libp2p-crypto": "^0.19.7",
     "multiaddr": "^10.0.1",
     "multihashes": "^4.0.3",
-    "peer-id": "^0.15.3",
+    "peer-id": "^0.16.0",
     "rlp": "^2.2.6",
     "strict-event-emitter-types": "^2.0.0",
     "varint": "^6.0.0"
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -386,7 +386,11 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   /**
    * Send TALKRESP message to requesting node
    */
-  public async sendTalkResp(remote: ENR | Multiaddr, requestId: RequestId, payload: Uint8Array): Promise<void> {
+  public async sendTalkResp(
+    remote: ENR | Multiaddr | INodeAddress,
+    requestId: RequestId,
+    payload: Uint8Array
+  ): Promise<void> {
     const msg = createTalkResponseMessage(requestId, payload);
     const nodeAddr = getNodeAddress(createNodeContact(remote));
     this.sendRpcResponse(nodeAddr, msg);
@@ -402,7 +406,7 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   /**
    * Sends a PING request to a node
    */
-  private sendPing(enr: ENR): void {
+  public sendPing(enr: ENR): void {
     this.sendRpcRequest({ contact: createNodeContact(enr), request: createPingMessage(this.enr.seq) });
   }
 
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -74,16 +74,16 @@ export interface IDiscv5CreateOptions {
  */
 export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   /**
+   * Session service that establishes sessions with peers
+   */
+  public sessionService: SessionService;
+
+  /**
    * Configuration
    */
   private config: IDiscv5Config;
 
   private started = false;
-
-  /**
-   * Session service that establishes sessions with peers
-   */
-  private sessionService: SessionService;
 
   /**
    * Storage of the ENR record for each node
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -637,9 +637,14 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
 
   // process events from the session service
 
-  private onEstablished = (enr: ENR, direction: ConnectionDirection): void => {
-    // Ignore sessions with non-contactable ENRs
-    if (!enr.getLocationMultiaddr("udp")) {
+  private onEstablished = (
+    nodeAddr: INodeAddress,
+    enr: ENR,
+    direction: ConnectionDirection,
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "eslint --color --ext .ts src/",
     "test": "yarn test:unit && yarn test:e2e",
     "test:unit": "mocha 'test/unit/**/*.test.ts'",
-    "test:e2e": "mocha 'test/unit/**/*.test.ts'"
+    "test:e2e": "mocha 'test/e2e/**/*.test.ts'"
```
dapplion, line 16: :pray:

acolytec3:
This should be ready for review.

src/service/service.ts
```diff
@@ -386,7 +386,11 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   /**
    * Send TALKRESP message to requesting node
    */
-  public async sendTalkResp(remote: ENR | Multiaddr, requestId: RequestId, payload: Uint8Array): Promise<void> {
+  public async sendTalkResp(
+    remote: ENR | Multiaddr | INodeAddress,
```
wemeetagain, line 390: Can you change this to only allow `nodeAddr: INodeAddress`?

And revert the changes to `createNodeContact`?

src/service/service.ts
```diff
@@ -402,7 +406,7 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   /**
    * Sends a PING request to a node
    */
-  private sendPing(enr: ENR): void {
+  public sendPing(enr: ENR): void {
```
wemeetagain, line 409: Can you change the param to `remote: ENR | Multiaddr`? Would allow pinging a peer via full multiaddr.

wemeetagain (changes requested):
Does it make sense to also also allow outbound unverified sessions?
They seem less useful, but you could dial a peer via a full multiaddr, and currently, if they respond with 'unverified' ENR (wrong or no IP+port), the session is dropped.
If we allowed unverified outbound sessions, we'd be able to use them for TALKREQ/RESP at least.

src/service/service.ts
acolytec3: To make sure I'm tracking, you want `remote : INodeAddress`?  If so, don't we need to update the function signature of `sendRpcResponse`  as well since it wants a `NodeContact` and not just the `INodeAddress`.  The challenge this gives for Portal Network (unless I'm missing something) is that we may not have access to a node's public key, just an invalid ENR that contains the nodeId.  We can use this `nodeId` plus the observed address to still specify which session to use but there's no clear way to get the full Multiaddr with an inlined public key from that since as far as I understand you can't go from a nodeId back to a public key.

acolytec3:
> Does it make sense to also also allow outbound unverified sessions? They seem less useful, but you could dial a peer via a full multiaddr, and currently, if they respond with 'unverified' ENR (wrong or no IP+port), the session is dropped. If we allowed unverified outbound sessions, we'd be able to use them for TALKREQ/RESP at least.

Potentially, though at least in a Portal Network context, we have no inbuilt way to obtain a full `Multiaddr` (including the public key).  We do have one subprotocol that relies entirely upon `TALKREQ` and drops the corresponding TALKRESPs and it could be useful to allow that subprotocol to work based just upon the INodeAddress.

acolytec3:
And now that I think about it, another way to approach this might be to have the sessionService "established" event include the observed socket address along with the ENR as well as a "verified" flag and then update the routing table to only if the verified is set.  Then, Portal Network can listen for that and update our own ENR cache (since we are already also maintaining our own routing tables) with the ENRs and multiaddrs and then we can call `TALKREQ/TALKRESP` with ENR or multiaddr.  Thoughts?

src/service/service.ts
acolytec3: Disregard.  I was looking at the wrong method.  This feedback should now be addressed.

acolytec3:
Disregard my previous feedback, I think this is now ready for review again.

acolytec3:
> Does it make sense to also also allow outbound unverified sessions? They seem less useful, but you could dial a peer via a full multiaddr, and currently, if they respond with 'unverified' ENR (wrong or no IP+port), the session is dropped. If we allowed unverified outbound sessions, we'd be able to use them for TALKREQ/RESP at least.
> 

I went ahead and added this in as I think I now can see where it can be useful down the line.  I can't think of a use case where we'll be able to make use of it within Portal Network in our current code base unless we set up an ENR/socket address cache since we wouldn't have any other way to know the correct socket address to send a TALKREQ to but as I noted above, we do have one use case where we might want to still send outbound messages to a node with an invalid ENR so this will be a useful stepping stone to that use case.

src/session/service.ts
```diff
@@ -510,28 +507,35 @@ export class SessionService extends (EventEmitter as { new (): StrictEventEmitte
...
-      // Verify the ENR is valid
-      if (this.verifyEnr(enr, nodeAddr)) {
-        // Session is valid
-        // Notify the application
+      // Verify the ENR endpoint matches observed node address
+      const verified = this.verifyEnr(enr, nodeAddr);
+
+      // Drop session if invalid ENR and session service not configured to allow unverified sessions
+      if (!verified && !this.config.allowUnverifiedSessions) {
+        log("ENR contains invalid socket address. Dropping session with %o", nodeAddr);
```
wemeetagain, line 515: I think there should be a `this.failSession(nodeAddr, RequestErrorType.InvalidRemoteENR, true)` after this log and the similar log below.

wemeetagain (changes requested):
> And now that I think about it, another way to approach this might be to have the sessionService "established" event include the observed socket address along with the ENR as well as a "verified" flag and then update the routing table to only if the verified is set. Then, Portal Network can listen for that and update our own ENR cache (since we are already also maintaining our own routing tables) with the ENRs and multiaddrs and then we can call `TALKREQ/TALKRESP` with ENR or multiaddr. Thoughts?

I think thats a good idea. We can update the `established` event type to:
`established: (nodeAddr: INodeAddress, enr: ENR, connectionDirection: ConnectionDirection, verified: boolean) => void;`
and filter out unverified sessions as a first step in `onEstablished`.

This also wouldn't effect performance/behavior in the default configuration, since only verified sessions will emit `established` events.

src/session/service.ts
acolytec3: Added.

acolytec3:
> I think thats a good idea. We can update the established event type to:
established: (nodeAddr: INodeAddress, enr: ENR, connectionDirection: ConnectionDirection, verified: boolean) => void;
and filter out unverified sessions as a first step in onEstablished.

I've made these updates.  Can we also bubble up this `established` event to the discv5 service layer?  I'd have to add a new event at that level.  Otherwise, I can hack around it with a little `;(discv5 as any).sessionService.on("established...` in our portalnetwork code if you don't want it exposed in the main API.

wemeetagain:
> > I think thats a good idea. We can update the established event type to:
> > established: (nodeAddr: INodeAddress, enr: ENR, connectionDirection: ConnectionDirection, verified: boolean) => void;
> > and filter out unverified sessions as a first step in onEstablished.
> 
> I've made these updates. Can we also bubble up this `established` event to the discv5 service layer? I'd have to add a new event at that level. Otherwise, I can hack around it with a little `;(discv5 as any).sessionService.on("established...` in our portalnetwork code if you don't want it exposed in the main API.

Maybe we can just make the `sessionService` public. You could then subscribe to events there without having to cast to any. Make it up to consumers of the library to not meddle with the sessionService in sketchy ways.

wemeetagain (approved):
Awesome! this will be a nice feature

PR #184, opened 2022-06-09
feat!: convert to esm only

wemeetagain:
- Upgrade libp2p dependencies
  - Includes updating the libp2p peer discovery layer
- Use ESM only
- Lint test files

### Migration help:

For consuming projects, there are two options:
1. If you upgrade your library to use ESM, you can import discv5 using `import ... from ...` syntax.
eg:
```ts
// using es modules
import { Discv5 } from "@chainsafe/discv5";

// use `Discv5`
```

3. If your project is using commonjs imports, you will need to use `await import` syntax to import discv5.
eg:
```ts
// using commonjs

const { Discv5 } = await import("@chainsafe/discv5")

// use `Discv5`
```


dapplion (commented):
Overall looks good to me!

dapplion (approved):
LGTM

PR #185, opened 2022-06-09
Run linter on tests

dapplion:


wemeetagain:
closing in favor of #184 which does the same

PR #192, opened 2022-06-24
feat: add `packet` to package.json `exports

acolytec3:
Adds `packet` to the `exports` field in package.json as we're using this in Ultralight and don't seem to be able to import it properly without hacking the package.json in node_modules after install

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,9 @@
     },
     "./libp2p": {
       "import": "./lib/libp2p/index.js"
+    },
+    "./packet": {
+      "import": "./lib/packet/index.js"
     }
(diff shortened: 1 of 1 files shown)


wemeetagain:
Can you update the commit message, see: https://github.com/ChainSafe/discv5/runs/7050634743?check_suite_focus=true#step:2:12

acolytec3:
> Can you update the commit message, see: https://github.com/ChainSafe/discv5/runs/7050634743?check_suite_focus=true#step:2:12

Done!

wemeetagain (approved):

PR #197, opened 2022-07-04
feat: upgrade crypto from bcrypto to noble

acolytec3:
- Replaces the `bcrypto` dependency with a combination of `@noble/hashes` and `@noble/secp256k1` (both are audited, modern, JS only ESM modules with no external dependencies)
- Removes `bcrypto` types
- Adds `lint:fix` script because `eslint` is your friend
- Selectively disabled eslint rules where necessary (mainly in tests where typescript would otherwise complain)


mpetrunic (commented):
I would very much love to drop bcrypto everywhere 😄

src/enr/v4.ts
```diff
@@ -1,31 +1,38 @@
...
 }
 
 export function createPrivateKey(): Buffer {
-  return secp256k1.privateKeyGenerate();
+  return Buffer.from(secp256k1.utils.randomPrivateKey());
 }
 
 export function publicKey(privKey: Buffer): Buffer {
-  return secp256k1.publicKeyCreate(privKey);
+  return Buffer.from(secp256k1.getPublicKey(Uint8Array.from(privKey), true));
```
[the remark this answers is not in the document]
acolytec3: Ah, didn't realize this.  Will check the other places where I have this and verify we aren't breaking anything.  In earlier attempts, the `noble` libraries were silently failing in some places where I passed the wrong input in.

acolytec3:
Okay, all the unit tests are passing now.  Not sure why the e2e tests are failing since I don't run the bootnodes so not sure if they're just not up or if something is still wrong somewhere in the code.  I'm going to test this some against ultralight and see if it works.

Would be curious for any feedback.

src/enr/enr.ts
```diff
@@ -350,6 +350,6 @@ export class ENR extends Map<ENRKey, ENRValue> {
     return encoded;
   }
   encodeTxt(privateKey?: Buffer): string {
```
wemeetagain, line 352: I'd prefer to remove `Buffer` entirely. Iirc, the only reason we needed it (vs vanilla `Uint8Array`) was bc of bcrypto.
Maybe we can just do this as a separate PR (it's also a breaking change)

src/enr/v4.ts
```diff
@@ -1,31 +1,39 @@
...
-import secp256k1 from "bcrypto/lib/secp256k1.js";
+import { keccak_256 as keccak } from "@noble/hashes/sha3";
+import * as secp256k1 from "@noble/secp256k1";
+import { hmac } from "@noble/hashes/hmac";
+import { sha256 } from "@noble/hashes/sha256";
+secp256k1.utils.hmacSha256Sync = (key: Uint8Array, ...msgs: Uint8Array[]) => {
+  const h = hmac.create(sha256, key);
+  msgs.forEach((msg) => h.update(msg));
+  return h.digest();
+};
```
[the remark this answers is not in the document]
acolytec3: Yup, sorry, was having trouble with that before but figured it it out.

acolytec3:
Added another commit to replace the noble re-export of node's `crypto` module with with the direct `crypto` import from nodejs.  Hopefully, this will allow me to properly polyfill it in our browser implementation.

paulmillr:
`base64url` could also be replaced with audited version of one inside https://github.com/paulmillr/scure-base.

src/enr/enr.ts
acolytec3: That feels like a pretty big lift to include in what is already not a small PR.  Are you good with differing that to if/when we get this one merged?

acolytec3:
> `base64url` could also be replaced with audited version of one inside https://github.com/paulmillr/scure-base.

We could put that in a second PR as well.  I'm guessing that's non-breaking and this one is already pretty big.

dapplion (changes requested):
Thank you for the contribution! Overall looks good

But please you or @wemeetagain must add benchmarks to show performance of hotpath ops vs this PR and current master.

acolytec3:
> Thank you for the contribution! Overall looks good
> 
> But please you or @wemeetagain must add benchmarks to show performance of hotpath ops vs this PR and current master.

Once I solve the remaining bugs, I can take a look at it.  This should definitely not be merged until the integration tests are passing as there's something wrong at the network layer (seems most likely to me to be related to the cipher code but not sure yet).

src/util/crypto.ts
```diff
@@ -11,13 +11,13 @@ secp256k1.utils.hmacSha256Sync = (key: Uint8Array, ...msgs: Uint8Array[]) => {
...
-export async function aesCtrEncrypt(key: Buffer, iv: Buffer, pt: Buffer): Promise<Buffer> {
+export function aesCtrEncrypt(key: Buffer, iv: Buffer, pt: Buffer): Buffer {
   const ctx = Crypto.createCipheriv("aes-128-gcm", key, iv);
   ctx.update(pt);
   return ctx.final();
 }
 
-export async function aesCtrDecrypt(key: Buffer, iv: Buffer, pt: Buffer): Promise<Buffer> {
+export function aesCtrDecrypt(key: Buffer, iv: Buffer, pt: Buffer): Buffer {
   const ctx = Crypto.createDecipheriv("aes-128-gcm", key, iv);
```
paulmillr, line 21: still aes-gcm here instead of aes-ctr

acolytec3: Ah, thanks.  That's an unused utility but will fix as well.

acolytec3: Actually, it's only aes-ctr in the unauthenticated bits of the handshake and then switches to aes-gcm for authenticated sessions per [here](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#node-discovery-protocol-v5---wire-protocol)

paulmillr:
The pure results on M1 are as such:

```
getPublicKey(utils.randomPrivateKey()) x 6,300 ops/sec @ 158μs/op; not 2896ops/sec
sign x 4,888 ops/sec @ 204μs/op
verify x 950 ops/sec @ 1ms/op
```

Your benchmark also benchmarks other stuff, like `Buffer.from(hex)`.

acolytec3:
> The pure results on M1 are as such:
> 
> ```
> generatePrivateKey() x 265,957 ops/sec @ 3μs/op
> getPublicKey(generatePrivateKey()) x 6,300 ops/sec @ 158μs/op
> sign x 4,888 ops/sec @ 204μs/op
> verify x 950 ops/sec @ 1ms/op
> ```
> 
> Your benchmark also benchmarks other stuff, like `Buffer.from(hex)`.

Yup, I'll leave it to @dapplion and @wemeetagain to decide whether to proceed. The only other optimization we can do would be to remove `Buffer` I think and I'm not sure it'll give that much of an improvement.

paulmillr:
yeah, i'm just saying that even without buffers, your abstraction layer may be too thick, because verification is ~1k ops/sec, not ~500, so it's useful to compare those raw results; they are totally achievable. Some thinking needs to be applied to what are the bottlenecks and at which process stages.

wemeetagain:
Haven't had a chance to dig into the benchmarks, but one thing we may be able to do re: the slowdown is make the crypto "pluggable", like we did with noise https://github.com/ChainSafe/js-libp2p-noise/pull/133

Then, for node applications that require highest possible performance, bcrypto or other native implementation can be passed in as configuration. And for browser applications, noble can be passed in.

acolytec3:
> Haven't had a chance to dig into the benchmarks, but one thing we may be able to do re: the slowdown is make the crypto "pluggable", like we did with noise [ChainSafe/js-libp2p-noise#133](https://github.com/ChainSafe/js-libp2p-noise/pull/133)
> 
> Then, for node applications that require highest possible performance, bcrypto or other native implementation can be passed in as configuration. And for browser applications, noble can be passed in.

Hmm, that's an interesting idea, also how we did for the transport layer.  I'm game to rework things down that route, though we'll have to think through the best approach given that it's really 4 different underlying dependencies (secp256k1, sha256, keccak, aes ciphers).  To my way of thinking, the aes ciphers is the least challenging one since I just subbed in the node crypto library implementations which seem to be actually slightly faster based on the couple of benchmarks I ran.  And, those are simple to polyfill in browser.

wemeetagain:
> have to think through the best approach given that it's really 4 different underlying dependencies 

Good point! I'm hoping to avoid spreading this to the ENR layer (I'm hoping to pull it out into its own package at some point)

The way it was done in noise was to add some interface (in noise called `ICryptoInterface`) that looks like:
```ts
export interface ICryptoInterface {
  hashSHA256: (data: Uint8Array) => Uint8Array
  ...
}
```
Then something that fulfills that interface gets passed in when the main object is constructed.

I guess some decisions could be made about which functions are synchronous vs asynchronous. And which crypto methods need / should be pluggable.

acolytec3:
> > have to think through the best approach given that it's really 4 different underlying dependencies
> 
> Good point! I'm hoping to avoid spreading this to the ENR layer (I'm hoping to pull it out into its own package at some point)
> 
> The way it was done in noise was to add some interface (in noise called `ICryptoInterface`) that looks like:
> 
> ```ts
> export interface ICryptoInterface {
>   hashSHA256: (data: Uint8Array) => Uint8Array
>   ...
> }
> ```
> 
> Then something that fulfills that interface gets passed in when the main object is constructed.
> 
> I guess some decisions could be made about which functions are synchronous vs asynchronous. And which crypto methods need / should be pluggable.

Make sense.  I've already half-way done that by consolidating the secp export in the `util` directory.  Shouldn't be too much of a lift to take it one step further to a full interface.

Would you be good with just letting the aes ciphers stand on their own using the node standard library dependencies?  That would simplify things (and I think we can remove those `aesCtr` helpers unless you want to use them somewhere).

paulmillr:
For aes, we're using this approach in EF repo: https://github.com/ethereum/js-ethereum-cryptography/blob/master/src/aes.ts

It uses node or web browser native built-ins. I'm sure there is zero need in polyfilling AES when web browsers support it.

acolytec3:
> For aes, we're using this approach in EF repo: https://github.com/ethereum/js-ethereum-cryptography/blob/master/src/aes.ts
> 
> It uses node or web browser native built-ins. I'm sure there is zero need in polyfilling AES when web browsers support it.
I'll take a look.  The polyfill approach is nice because we only have to use nodejs builtins API and not maintain separate branches but it also does introduce that secondary dependency.  Will think on it as I work through implementing a new crypto interface for this stuff.

dapplion:
> I guess some decisions could be made about which functions are synchronous vs asynchronous

@wemeetagain please don't make the API async :pray: we are already doing enough Promises everywhere

acolytec3:
> > I guess some decisions could be made about which functions are synchronous vs asynchronous
> 
> @wemeetagain please don't make the API async :pray: we are already doing enough Promises everywhere

As long as it's followsc roughly what I've done all far in this PR, I shouldn't need to make anything else async that isn't already but will keep that in mind as I work on the interface.

acolytec3:
I'm starting to dig into an interface based approach to the session/keys/crypto stuff and this is kind of what I'm thinking.  I'd like to remove the `v4` file and `enrKeyPair` class entirely and leverage the existing `IKeypair` and/or `IKeypairClass` interfaces for all of the methods.  It seems like unnecessary layers of abstraction and duplication of methods to have verify/sign/hash methods in both `v4` and in the `keypair` classes.  I think it makes sense to have a single set of utility methods for the hashing (keccak and sha256 as applicable), and then a single keypair interface that each keypair type implements.

acolytec3:
@wemeetagain and I met today and talked through this PR and at this point, there doesn't seem like a viable path forward for implementing a crypto interface for the `secp256k1` code that won't also require a much deeper rewrite of the whole ENR implementation so I'm not going to continue more work on this specific PR going forward but will rather explore how to get Ultralight working with the existing master branch and try and solve the import issues related to `bcrypto` browser bindings in the near term.  I think there are some specific items that can be pulled out that would be fruitful next steps:
- Replace the bcrypto aes method calls with Nodejs builtins.  That should be relatively easy to pull into a standalone PR as it is isolated from the other crypto work contained herein.  
- Also, the `sha256` implementation could likely be switched for something else if desired (either your own [AS version](https://www.npmjs.com/package/@chainsafe/as-sha256) or the `noble/hashes` version used here) without too much work.
- Switching secp256k1 to an interface that would allow any conforming implementation to be brought would require a deeper rewrite though, likely entirely disposing of the existing ENR class and replacing it with a collection of static method calls that use some global secp25k1 interface that could be provided at build time by the consuming application.

acolytec3:
One more note on this.  I was finally able to get webpack to find the correct `bcrypto` bindings in the browser by adding a new environmental variable to explicitly tell `bcrypto` which binding to provide.  I'm not sure if there is some industry wide environmental variable named `NODE_BACKEND` but I found it [being used](https://github.com/bcoin-org/bcrypto/blob/a0e2635eb5bd50671703853391589c5bc428f51a/lib/keccak.js#L9) throughout the `bcrypto` code base and [telling webpack to explicitly set it to `js`](https://github.com/ethereumjs/ultralight/pull/328/commits/fd197980137f45e479b37b25b8f9fca221b6d88e) allows Ultralight to compile using the master branch and run in the browser without having issues locating the right bcrypto backend.

acolytec3:
I still vote to drop bcrypto at the earliest opportunity but that may be tricky with secp256k1 until you allow for an interface to be passed in.

acolytec3:
So I left this sitting as a relic in case I ever had the energy or inspiration to come back to it.  Are y'all still wedded to the idea of bcrypto for performance's sake?  This PR is purely prior art at this point but I'd still love to be done with bcrypto.  It causes the occasional annoying core dump and random test failures for Ultralight that I don't think are attributable to our code but I'm curious if the team would be willing to reconsider dropping bcrypto and going entirely to something more modern?  I'm not sure if the discv5 usage within lodestar is an area where you're concerned about performance bottlenecks but either switching entirely to `noble` variants or at least upgrading the various crypto usage to something like `bitcoin-ts` (for secp256k1) and the nodejs buitlins for aes ciphers (as suggested in #200) would eliminate the annoyances of bcrypto for us.  I'd be happy to take this on if y'all are willing.

acolytec3:
@wemeetagain @philknows ^^

acolytec3:
Going to close this as superseded by #302

Issue #201, opened 2022-09-20
External memory leak


tuyennhv:
**Description**
In lodestar if it sets "udp" port in enr, then external memory keeps increasing

<img width="804" alt="Screen Shot 2022-09-20 at 13 57 21" src="https://user-images.githubusercontent.com/10568965/191188440-7eb660a6-32cc-4c73-aa77-4c8914538fe6.png">

It's likely this issue links to the NODES handlers

<img width="1224" alt="Screen Shot 2022-09-20 at 13 58 22" src="https://user-images.githubusercontent.com/10568965/191188607-bd1aeb32-fbaa-40b7-86c0-1814677eaf9e.png">

This happens with discv5 `0.7.1` (could be with the latest version too)

tuyennhv:
I've had several tests at discv5:
- udp send/receive tests
- encode nodes response test
- async/await send api tests
- different improvements in handleFindNode

none of them work (either cannot reproduce the issue with the discv5 test only, or the issue still happens if I try with lodestar)

`discv5` may not be the root cause, the issue may happen because a side effect of setting enr to udp that cause lodestar receives so may peers and it switches peers a lot, see https://github.com/ChainSafe/lodestar/issues/4623#issuecomment-1374418235

tuyennhv:
closing the issue, root cause was https://github.com/ChainSafe/node-snappy-stream/issues/16

PR #202, opened 2022-09-21
fix: add async await

tuyennhv:
**Motivation**
- There is an external memory leak issue in #201 , it's very possible it links to the use of udp socket
- The `send` api of `udp` is async, it's natural to enforce async/await for consumer apis

**Description**
- The main change to me that possibly affect the external memory is to send NODES response one by one, instead of in parallel
- Modify eslint to enforce async/await everywhere

**TODOs**
- [x] Deploy to some nodes => deployed to feat2 novc, sm1v
- [ ] Monitor metrics in node

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
     "@typescript-eslint/no-use-before-define": "off",
     "@typescript-eslint/semi": "error",
     "@typescript-eslint/type-annotation-spacing": "error",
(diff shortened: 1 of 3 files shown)


tuyennhv:
this does not completely resolve the issue, however it helps reduce increasement of external memory to some degree. Below is a chart of 6-7 days

Screen Shot 2022-09-27 at 10 42 41

PR #206, opened 2022-10-20
Add IP based rate limiting

dapplion:
UDP based transports exposed to an un-trusted networks must have the ability limit packet inflow for safety.

This PR adds rate limiting based on:
- Global quota for all connections
- Single IPs

@wemeetagain Rate limiting must happen before decoding the packet. However the current architecture the transport decodes the packet before emitting it. So the rate limiting must happen inside the transport class. This makes it weird in the way the classes are injected on each other. Do you have any suggestion on where it's best to instantiate the RateLimiter class where it should live?


wemeetagain:
A few options:
- leave as is
- change Transport to emit `("packet", mu: Multiaddr, encodedPacket: Uint8Array)` instead of the decoded packet
  - move decoding and rate limiting to the layer above

wemeetagain:
marking as breaking since it breaks some of the exported interfaces that we know the ultralight folks are using.

wemeetagain (approved):

PR #214, opened 2022-10-29
feat: drop is-ip by representing IPs internally as octets

dapplion:
Current implementation coverts IPs from string and bytes multiple times unnecessarily. This motivated the need for is-ip and ip6addr libraries.

This PR:
- Represents PONG message IP:port combo as bytes internally
- Tally votes without multiaddr representation


wemeetagain:
@dapplion @tuyennhv  ready for review

wemeetagain:
this PR would be a lot nicer with a better name for the `IPUDP` type...

src/util/ip.ts
```diff
@@ -0,0 +1,122 @@
+import { multiaddr, Multiaddr } from "@multiformats/multiaddr";
+import { ENR } from "../enr/index.js";
+
+export type IP = { type: 4 | 6; octets: Uint8Array };
+export type IPUDP = IP & { udp: number };
```
dapplion, line 5: @wemeetagain I would do
```ts
interface SocketAddress {
  ip: IP
  port: number
}
```
Or if you really need to clarify it's UDP then
```ts
interface SocketAddressUDP {
  ip: IP
  udp: number
}
```
But I don't see the point, this is used only internally

wemeetagain: :+1: will refactor

src/message/encode.ts
```diff
@@ -56,17 +55,18 @@ export function encodePingMessage(m: IPingMessage): Buffer {
...
-  }
-  if (m.recipientPort < 0 || m.recipientPort > 65535) {
+  if (m.addr.port < 0 || m.addr.port > 65535) {
     throw new Error("invalid port for encoding");
   }
   return Buffer.concat([
     Buffer.from([MessageType.PONG]),
-    RLP.encode([toBuffer(m.id), toBuffer(m.enrSeq), tuple, m.recipientPort]),
+    RLP.encode([
+      //
```
tuyennhv, line 64: some comments here?

tuyennhv (approved):
this looks great, we should include this in the next version of lodestar

philknows:
For whoever is creating the PR to bump Lodestar to discv5 1.4.0, can you additionally add v1.3.0 milestone to it as well? Thanks!

src/message/encode.ts
dapplion: Added this to force prettier into placing the arguments vertically

src/util/ip.ts
```diff
@@ -0,0 +1,131 @@
...
+      break;
+  }
+}
+
+// Protocols https://github.com/multiformats/multiaddr/blob/master/protocols.csv
+// code  size  name
+// 4     32    ip4
+// 41    128   ip6
+// 273   16    udp
+const Protocol = {
```
dapplion, line 82: What was the issue with using an enum here?

wemeetagain: its used to simplify `multiaddrFromSocketAddress`. PR welcome if it bothers you.

PR #216, opened 2022-11-01
fix: add more checks to handleFindNode

jtraglia:
This PR fixes a few things in the FINDNODE request handler:

* We should filter out invalid/duplicate distances.
* According to [the spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#findnode-request-0x03), the recommended result limit is 16, not 15.
* When returning an empty response, `total` should be one.

### DIFF
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -743,7 +743,7 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
     for (const distance of new Set(distances)) {
       // filter out invalid distances
       if (distance < 0 || distance > 256) {
-        return;
+        continue;
       }
       // if the distance is 0, send our local ENR
       if (distance === 0) {
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -740,7 +740,12 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   private handleFindNode(nodeAddr: INodeAddress, message: IFindNodeMessage): void {
     const { id, distances } = message;
     let nodes: ENR[] = [];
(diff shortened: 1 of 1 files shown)


src/service/service.ts
```diff
@@ -740,7 +740,12 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
   private handleFindNode(nodeAddr: INodeAddress, message: IFindNodeMessage): void {
     const { id, distances } = message;
     let nodes: ENR[] = [];
-    distances.forEach((distance) => {
+    let distinctDistances = distances.filter((n, i) => distances.indexOf(n) === i);
```
wemeetagain, line 743: maybe better to just convert to a set to filter out duplicates
```ts
for (const distance of new Set(distances)) {
  // filter out invalid distances
  ...
```

wemeetagain:
> When returning an empty response, total should be one

Why is that?

src/service/service.ts
jtraglia: Oh that looks much better. Thank you. Will fix.

jtraglia:
> > When returning an empty response, total should be one
> 
> Why is that?

Well `total` is the number of packets being sent/received. If you're sending something, it must be at least one, even if the number of nodes is zero. For reassurance, see the following other implementations:

* https://github.com/sigp/discv5/blob/master/src/service.rs#L1006-L1014
* https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/v5_udp.go#L831-L833

wemeetagain (approved):
Thanks @jtraglia

looks good to me

PR #218, opened 2022-11-02
fix: throw error when encoding/decoding PONG with invalid port

jtraglia:
According to [the spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#pong-response-0x02), `recipient-port` is an (unsigned) 16-bit integer. This PR ensures that those port values are valid.

* Remove a commented out line of code, doesn't seem useful to keep.
* When decoding, check that the length of the field isn't greater than 2 bytes.
* When encoding, check that the number is in the valid port range.
* Add a test for encoding a PONG message with a 1-byte port.
* Add a new test function that handles invalid messages.


wemeetagain (approved):
:pray:

PR #219, opened 2022-11-02
feat: limit nodes response to 16 packets

jtraglia:
According to [the spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#nodes-response-0x04):

> Implementations may place a limit on the allowed maximum for total. If exceeded, additional responses may be ignored.

This PR implements this feature. It will ignore response packets after `MAX_NODES_TOTAL_PACKETS` (16). This should be more than enough, considering the FINDNODE request handler should limit the result to 16 nodes, which can be fit into 6 packets if packed correctly.

### DIFF
--- a/src/session/constants.ts
+++ b/src/session/constants.ts
@@ -1,0 +1,1 @@
+export const MAX_NODES_TOTAL_PACKETS = 16;
--- a/src/session/service.ts
+++ b/src/session/service.ts
@@ -42,6 +42,7 @@ import { getNodeAddress, INodeAddress, INodeContactType, nodeAddressToString, No
 import LRUCache from "lru-cache";
 import { TimeoutMap } from "../util/index.js";
 import { IDiscv5Metrics } from "../service/types.js";
+import { MAX_NODES_TOTAL_PACKETS } from "./constants.js";
 
 const log = debug("discv5:sessionService");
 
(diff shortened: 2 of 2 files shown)


wemeetagain:
Great!
Is this too lenient? Where did you get 16 from?

jtraglia:
> Where did you get 16 from?

Typically, 3 distances are requested in a FINDNODE request. Given that a bucket can contain at most 16 nodes and it's possible to fit 3 ENRs into a packet (can't guarantee 4) it would take 16 packets to return all of the nodes.

> Is this too lenient?

Maybe. The FINDNODE request handler **should** limit the request to 16 nodes (5.33, 6 packets rounded up).

Just to give you some ideas of what other clients are doing:

* Geth/Prysm [limits the number of responses to 5](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/v5_udp.go#L43).
  * Seems that they don't care about the 16th node 🤷
* Lighthouse has two limits: [15 packets](https://github.com/sigp/discv5/blob/master/src/service.rs#L644-L649) and [6 packets](https://github.com/sigp/discv5/blob/master/src/service.rs#L732-L735) by default.
  * It's a bit redundant. It's an interesting implementation.
* Teku [limits the number of packets to 16](https://github.com/ConsenSys/discovery/blob/f76b7eda64b434e37a5787b9cdbbb99e2a2d256b/src/main/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandler.java#L19).

I'm a fan of keeping it simple. To me, this is intended to prevent DOS attacks.

wemeetagain:
Thanks for the explanation

wemeetagain (approved):

PR #231, opened 2023-01-04
fix: cache the NodeAddressString

tuyennhv:
**Motivation**
- Even through the cost of computing NodeAddressString may not be a lot, it's redundant to keep computing it in the same flow
- An example of this is the "handleFindNodes" flow where we compute NodeAddressString 5 times:
  - Session service to compute NodeAddressString, then send to Discv5 service
  - Discv5 service find 16 nodes, with 4 node per packet it'd send 4 responses
  - For each response we compute NodeAddressString again

**Description**
- Cache NodeAddressString inside the `INodeAddress`


dapplion (approved):

wemeetagain:
Can you show the benefit in terms of CPU and/or memory savings?

PR #232, opened 2023-01-04
feat: chunkify handleFindNode

tuyennhv:
**Motivation**
- In handleFindNode, we keep adding value to the resulting array then do a `slice()` to make sure the resulting nodes contain 16 items at most, we should `push()` then `check()` instead
- The embeded chunkify work has duplicate code and not testable

**Description**
- Add chunkify util

### DIFF
--- a/src/service/service.ts
+++ b/src/service/service.ts
@@ -782,8 +782,8 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
    */
   private handleFindNode(nodeAddr: INodeAddress, message: IFindNodeMessage): void {
     const { id, distances } = message;
-    let nodes: ENR[] = [];
-    for (const distance of new Set(distances)) {
+    const nodes: ENR[] = [];
+    topLoop: for (const distance of new Set(distances)) {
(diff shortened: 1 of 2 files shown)


src/service/service.ts
```diff
@@ -795,33 +795,30 @@ export class Discv5 extends (EventEmitter as { new (): Discv5EventEmitter }) {
...
       } else {
-        nodes.push(...this.kbuckets.valuesOfDistance(distance));
-      }
-    }
-    // limit response to 16 nodes
-    nodes = nodes.slice(0, 16);
-    if (nodes.length === 0) {
-      log("Sending empty NODES response to %o", nodeAddr);
-      try {
-        this.sessionService.sendResponse(nodeAddr, createNodesMessage(id, 1, nodes));
```
wemeetagain, line 807: After this PR we no longer send a NODES response if we've found no nodes, is that intended?

tuyennhv: when there is no nodes, the `chunkify` function returns `[[]]` (see the unit test), in that case we still send 1 response with total = 1

PR #233, opened 2023-01-13
feat!: switch to metrics generator pattern

### LINKED ISSUE #225: Switch to metrics generator pattern
See https://github.com/ChainSafe/lodestar/pull/4737#issuecomment-1311438492
This library should be able to register its own metrics rather than rely on consumer to register its metrics.

As a rough example see: https://github.com/libp2p/js-libp2p-tcp/pull/223

wemeetagain:
Resolves #225 

This will allow us to add metrics in a more agile way, since this library now maintains all code to create metrics.

This is a breaking change, since now, to enable metrics, instead of passing a `IDiscv5Metrics` object in the constructor, you pass a `Registry` object (and this library constructs the `IDiscv5Metrics` object)

Before:
`Discv5.create({..., metrics})`

After:
`Discv5.create({..., metricsRegistry})`

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "debug": "^4.3.1",
     "dgram": "^1.0.1",
     "err-code": "^3.0.1",
+    "prom-client": "^14.1.0",
     "rlp": "^2.2.6",
     "strict-event-emitter-types": "^2.0.0",
     "varint": "^6.0.0"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -105,6 +105,7 @@
     "debug": "^4.3.1",
     "dgram": "^1.0.1",
     "err-code": "^3.0.1",
+    "prom-client": "^14.1.0",
```
dapplion, line 108: I don't think discv5 should depend on prom-client directly. Instead just expect an specific interface like gossipsub

dapplion (approved):

PR #241, opened 2023-01-26
feat!: refactor ENR

### LINKED ISSUE #240: ENR API is error-prone
ENR API and types are error prone since it handles both mutable and immutable ENRs, and mixing these behaviors makes the types lax and adds confusing behavior. Eg: `encode` takes an optional private key. Encoding can fail if the ENR has been mutated and not resigned. Etc.

A better design would prevent error cases that currently have to be handled.

eg:
```ts
class ImmutableENR {
  set(..) {
    throw
  }
  delete(..) {
    throw
  }
  encode(): Uint8Array {
    // can't throw
  }
}

class MutableENR {
  privateKey: Uint8Array;
  encode(): Uint8Array {
    // can't throw (except in truly exceptional case)
    sign(..., this.privateKey);
    ...
  }
```

wemeetagain:
Refactor the ENR class to resolve #240 

Now instead of a single `ENR` class, there are two classes: `ENR` and `SignableENR`.

- `ENR` is an immutable object (enforced by type, no run-time check). enr buffers received from the network, enr strings passed in to the program, etc. should be decoded using `ENR.decode` or `ENR.decodeTxt`. There is no ability to set new key-values.
- `SignableENR` is an enr that includes a private key, and is able to be updated (and signed / re-signed).
  - It may be either re-created by decoding `SignableENR.decode(buf, keypair)`.
  - New enrs can be created with `new SignableENR`, `SignableENR.createvV4(keypair)` or `SignableENR.createFromPeerId(peerId)`.
  - key-values may be updated by `enr.set(key, value)` as well as setters for `enr.ip`, `enr.ip6`, `enr.tcp`, `enr.tcp6`, `enr.udp`, `enr.udp6`
  - a `SignableENR` can be converted to ENR with `enr.toENR()`. the resultant enr is a copy of the `SignableENR` with its current values.
- Both `ENR` and `SignableENR` extend `BaseENR`, which gives the following important properties / getters:
  - `kvs: ReadonlyMap<ENRKey, ENRValue>` - use this to access raw key-values
  - `signature` - the signature of the enr
    - in `ENR` this should never change
    - in `SignableENR`, this may trigger a re-sign if kvs has been updated.
  - `seq` - the sequence number
  - `nodeId` - the node id as a hex string
  - `publicKey` - the public key as a `Uint8Array`
- both `ENR` and `SignableENR` have `toObject` methods and `fromObject` static methods for cheaper structured cloning

Migration notes:
- replace `enr.get(k, v)` with `enr.kvs.get(k, v)`
- replace `ENR.createV4(pubkey)` with `SignableENR.createV4(keypair)` (use `createKeypair` to create a `Keypair` object)
- if a `SignableENR` needs to be converted to an `ENR`, use `enr.toENR()` which will create a copy of the current ENR with the current values.


dapplion (approved):
Looks great! Thanks for the neat refactor

PR #244, opened 2023-03-04
Add c8 based coverage reports for e2e and unit tests

0xTylerHolmes:
This PR adds c8 based coverage reports

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,0 +1,20 @@
+{
+  "all": true,
+  "cache": false,
+  "extension": [".ts", ".js", ".jsx", ".tsx"],
+  "exclude": [
+    "**/node_modules/**",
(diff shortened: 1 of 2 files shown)


dapplion:
Should the coverage report be acted upon? i.e. upload to some remote tracker

0xTylerHolmes:
I left it as is just to gain some quick metrics and keep it simple. 

If you want I could look into a good way to automate this and make another PR.

dapplion:
> I left it as is just to gain some quick metrics and keep it simple.
> 
> If you want I could look into a good way to automate this and make another PR.

CC @wemeetagain

wemeetagain:
looks good, we just need to update the c8 version (now at 10.x)

PR #245, opened 2023-04-28
feat!: add IPv6 support

### LINKED ISSUE #167: Support ipv6
I tried to use this library in an environment where ipv6 and ipv4 co-exist, but encountered some problems:

1. In `UDPTransportService`, the socket object is created according to the `multiaddr.toOptions().family`, which may be `udp4` or `udp6`. When the local socket object is `udp4` but the remote host is an ipv6 address, calling `this.socket.send` will return an error:

```
Error: send EINVAL aaaa:aaaa:aaaa:aaaa:aaaa:aaaa:aaaa:aaaa
    at doSend (dgram.js:692:16)
    at defaultTriggerAsyncIdScope (internal/async_hooks.js:430:12)
    at afterDns (dgram.js:638:5)
    at processTicksAndRejections (internal/process/task_queues.js:81:21) {
  errno: -22,
  code: 'EINVAL',
  syscall: 'send',
  address: 'aaaa:aaaa:aaaa:aaaa:aaaa:aaaa:aaaa:aaaa',
  port: 12345
}
```

Maybe we should create and bind two socket objects `udp4` and `udp6`, and then judge which socket object to use when sending?

2. In many places in the code, `enr.getLocationMultiaddr("udp")` is used directly without judgment

For example, [this](https://github.com/ChainSafe/discv5/blob/5ed517d2a99eb43c59542cfd44df5634590f0940/src/service/service.ts#L816) place should determine whether `message.recipientIp` is an ipv4 address or an ipv6 address before calling `enr.getLocationMultiaddr("udp")`. If `message.recipientIp` is an ipv6 address, but `currentAddr` is an ipv4 address, the two cannot be equal, which will cause the local node to keep increasing its own `enr.seq`

And other places, like [here](https://github.com/ChainSafe/discv5/blob/5ed517d2a99eb43c59542cfd44df5634590f0940/src/service/service.ts#L373) and [here](https://github.com/ChainSafe/discv5/blob/5ed517d2a99eb43c59542cfd44df5634590f0940/src/service/service.ts#L421) and [here](https://github.com/ChainSafe/discv5/blob/5ed517d2a99eb43c59542cfd44df5634590f0940/src/service/service.ts#L640) (Although I don't know what these places are doing)

wemeetagain:
Resolves #167

- Adds the capability to specify either an ip4, ip6, or both bind addresses instead of just a single bind address
- By specifying the addresses to bind to, this sets the `IPMode` (either ip4, ip6, or both).
- Based on the specified `IPMode`, several things happen:
  - a udp4 socket and/or a udp6 socket are opened
  - discovered peers will be contacted based on the proper entry in their enr (either ip+udp or ip6+udp6) (ip6 is the default in the case of the 'both' IPMode) using the proper socket

### DIFF
--- a/src/transport/types.ts
+++ b/src/transport/types.ts
@@ -3,6 +3,8 @@ import StrictEventEmitter from "strict-event-emitter-types";
 import { Multiaddr } from "@multiformats/multiaddr";
 
 import { IPacket } from "../packet/index.js";
+import { BaseENR } from "../enr/enr.js";
+import { SocketAddress } from "../util/ip.js";
 
 export interface ISocketAddr {
   port: number;
@@ -24,11 +26,20 @@ export interface ITransportEvents {
 }
 export type TransportEventEmitter = StrictEventEmitter<EventEmitter, ITransportEvents>;
 
+export type IPMode = {
+  ip4: boolean;
+  ip6: boolean;
+};
+
 export interface ITransportService extends TransportEventEmitter {
-  multiaddr: Multiaddr;
+  bindAddrs: Multiaddr[];
+  ipMode: IPMode;
+
   start(): Promise<void>;
   stop(): Promise<void>;
   send(to: Multiaddr, toId: string, packet: IPacket): Promise<void>;
+
+  getContactableAddr(enr: BaseENR): SocketAddress | undefined;
 
   /** Add 1 expected response of unknown length from IP for rate limiter */
   addExpectedResponse?(ipAddress: string): void;
--- a/src/transport/types.ts
+++ b/src/transport/types.ts
@@ -26,10 +26,33 @@ export interface ITransportEvents {
 }
 export type TransportEventEmitter = StrictEventEmitter<EventEmitter, ITransportEvents>;
 
-export type IPMode = {
-  ip4: boolean;
-  ip6: boolean;
-};
+export type IPMode =
+  | {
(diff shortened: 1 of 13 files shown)


src/transport/types.ts
```diff
@@ -24,12 +26,21 @@ export interface ITransportEvents {
 }
 export type TransportEventEmitter = StrictEventEmitter<EventEmitter, ITransportEvents>;
 
+export type IPMode = {
+  ip4: boolean;
+  ip6: boolean;
```
dapplion, line 31: Can it support both ip4 and ip6? Can it support none? Intuitively seems there are three possible options: ip4, ip4&ip6, ip6. Maybe it's best to have an enum (at least internally) to represent the 3 possible states?

test/e2e/mainnetBootnodes.test.ts
```diff
@@ -25,6 +25,9 @@ describe("discv5 integration test", function () {
       ip6: multiaddr(`/ip6/::/udp/${port++}`),
     },
   ]) {
+    // ip6 test fails in github runner
+    if (process.env.CI && bindAddrs.ip6) return;
```
dapplion, line 29: If this test is run in mocha, it's best to do `this.skip()` to signal test existence on all suites

dapplion (approved):

Issue #252, opened 2023-08-10
`lru-cache` is devDependency


acolytec3:
In upgrading some dependencies for Ultralight, I noticed some issues related to `lru-cache` in `discv5`. I'm not 100% sure this is the source of our issues but this dependency is directly imported in the `session/service.ts` file but is only declared as a devDependency in package.json

Would you be open to a PR moving this to the dependencies property so we can see it properly within Ultralight's dependency tree?

acolytec3:
Resolved by #253

PR #253, opened 2023-08-10
fix: Add lru-cache to deps

acolytec3:
Adds `lru-cache` to dependencies so it can be found in library consumer dependency trees.


acolytec3:
Fixes #252

acolytec3:
@wemeetagain this should be ready to go

wemeetagain (approved):

PR #262, opened 2023-09-11
chore: remove dgram from package.json

acolytec3:
Removes `dgram` as a dependency as this is a core Nodejs module now.  Will also keep this npm warning message from appearing when installing dependencies for discv5.

`npm WARN deprecated dgram@1.0.1: npm is holding this package for security reasons. As it's a core Node module, we will not transfer it over to other users. You may safely remove the package from your dependencies.`

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "@types/eslint": "^6.1.3",
     "@types/lru-cache": "^5.1.1",
     "@types/mocha": "^8.0.3",
-    "@types/node": "^12.0.10",
+    "@types/node": "^12.20.55",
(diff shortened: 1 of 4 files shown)


wemeetagain:
Can you also replace the line `import ... from "dgram"` with `import ... from "node:dgram"`?

acolytec3:
> Can you also replace the line `import ... from "dgram"` with `import ... from "node:dgram"`?

Done!

nflaig:
@types/node is still on version 12, should update that to 20 which then also has the types for dgram

wemeetagain (approved):
LGTM

Issue #265, opened 2023-09-22
`UnhandledPromiseRejection` in `sendPing`


acolytec3:
We've recently started noticing some unhandled promise rejections that are being thrown by `discv5.sendPing` inside of Ultralight.  The stack trace for the error looks like below:
```js
    at callback (file:///home/jim/development/ultralight/node_modules/@chainsafe/discv5/lib/service/service.js:354:33)
    at SessionService.rpcFailure (file:///home/jim/development/ultralight/node_modules/@chainsafe/discv5/lib/service/service.js:811:13)
    at SessionService.emit (node:events:514:28)
    at SessionService.failRequest (file:///home/jim/development/ultralight/node_modules/@chainsafe/discv5/lib/session/service.js:650:14)
    at SessionService.handleRequestTimeout (file:///home/jim/development/ultralight/node_modules/@chainsafe/discv5/lib/session/service.js:615:18)
    at TimeoutMap.onTimeout (file:///home/jim/development/ultralight/node_modules/@chainsafe/discv5/lib/session/service.js:98:84)
    at Timeout.<anonymous> (file:///home/jim/development/ultralight/node_modules/@chainsafe/discv5/lib/util/timeoutMap.js:25:22)
    at listOnTimeout (node:internal/timers:569:17)
    at process.processTimers (node:internal/timers:512:7)
```

and the debugger links the source of the error to [this promise rejection in `sendPing`](https://github.com/ChainSafe/discv5/blob/3c4534c3a8c07c71317b92f7677b14ce191ab1a5/src/service/service.ts#L466).  

I looked through the calls to `sendPing` in the discv5 codebase and it doesn't look like any of them are directly wrapped in `try`/`catch` blocks so not sure where I need to look to see why this isn't being handled since it looks like it's a basic session timeout issue (which shouldn't be a critical requiring a consuming application to handle it.

Since we never call `discv5.sendPing` directly inside Ultralight, I'm trying to determine if this is some new issue that has cropped up with `discv5` or if it's something Ultralight did (or didn't) do with regard to properly guarding network calls with our `discv5` client.  

Any suggestions?

For our part, we're temporarily employing a `process.on('uncaughtException...` handler to keep Ultralight from crashing for now since as best I can tell this comes from trying to ping unreachable nodes and the rejections just not being handled properly.

philknows:
Resolved by #284

PR #284, opened 2024-01-19
fix: improve error handling

wemeetagain:
## Motivation
- https://github.com/ChainSafe/discv5/issues/265
- https://github.com/ChainSafe/lodestar/issues/6320

## Notes
- refactor callback handling
  - move from callback style `(err, resp) => void` to deferred promise style `{resolve: (err) => void, reject: (err) => void}`
  - ensure callbacks are called in each rpc handler and in each error case
- remove hanging promises
  - add linter rule to ensure nothing remaining
- move error handling of node contacts into `getNodeContact` rather than the `getNodeAddress` "getter"
  - this makes error handling much more predictable, throwing upfront, rather than waiting for a later event to trigger an error
- wrap rpc handlers in try/catch statements
  - ensure that logic handed in rpc request and response handlers are unable to trigger unhandled exceptions

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -48,6 +48,7 @@ module.exports = {
     "prefer-const": "error",
     "quotes": ["error", "double"],
     "@chainsafe/node/file-extension-in-import": ["error", "always", {esm: true}],
+    "@typescript-eslint/no-floating-promises": "error",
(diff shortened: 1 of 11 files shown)


nflaig (approved):
lgtm

PR #285, opened 2024-01-24
feat: use ethereum-cryptography for enr crypto

wemeetagain:
In order to make `@chainsafe/enr` more usable as a standalone library, we'd like to remove its heavy dependency `bcrypto` and replace it with something more suitable for cross-environment usage, `ethereum-cryptography`.

For `@chainsafe/discv5`, which already uses `bcrypto`, we can continue to use its enr crypto. This PR adds a side effect of switching enr crypto to use `bcrypto` when importing from `@chainsfafe/discv5`.

For users of `@chainsafe/discv5` who would like to use `ethereum-cryptography` for enr crypto, you can always override the enr crypto like so:
```ts
import {setV4Crypto, defaultCrypto} from "@chainsafe/enr";

setV4Crypto(defaultCrypto)
```

### DIFF
--- a/packages/discv5/README.md
+++ b/packages/discv5/README.md
@@ -39,6 +39,17 @@ const libp2p = new Libp2p({
 
 ```
 
+## Additional features
+
+#### Faster nodejs ENR crypto
+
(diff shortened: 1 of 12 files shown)


nflaig (approved):
lgtm

PR #302, opened 2025-01-03
feat: replace bcrypto

### LINKED ISSUE #301: Drop `bcrypto`
I have attempted this here #197  but it didn't make sense at the time because of the performance downsides.  It's coming around again for me because Ultralight still sees occasional OOM and coredump errors from `bcrypto` (both in our CI and even occasionally in our live nodes) and it's kind of a bummer to be stuck using an old-ish heavy C library that keeps causing us issues with no easy way to remediate.  One suggestion that @wemeetagain made at the time was to implement a `crypto` interface that would allow any user to bring their own crypto library and implement wrapper code to make it work with the given interface.  

We did something similar in `ethereumjs` a while back with the direct benefit that our code natively uses `noble/crypto` for our crypto use cases but allows the use of something like `@polkadot/wasm-crypto` that was 2-10x faster (but also not dependent on old C code).

From a review of the code, a working PR would look something like this:

A new crypto interface object that exposes
- `keccak` - has an implementation in both `noble` and `wasm-crypto`
- `secp256k1`- has an implementation in both `noble` and `wasm-crypto`
- `sha256`- has an implementation in both `noble` and `wasm-crypto`
- `hdkf` - has an implementation in `noble` for sure - would have to research if wasm-crypto provides this (or we could find another suitable alternative)
- `randomBytes` - `noble` has an implementation of this - not sure performance wise if you need anything else

Use the nodejs builtins for the AES cipher stuff (already done in #197).

I would be willing to pick this up if y'all have an appetite for it.

acolytec3:
This is a redo of #197 and fixes #301.

Problem: `discv5` uses `bcrypto` dependency for performant `secp256k1` and `sha256` functions.  This library causes occasional CI failures in ultralight due to in explicable OOMs and other indecipherable C++ errors

Solution: Replace `bcrypto` usage with a `discv5Crypto` interface similar to the `v4Crypto` interface exposed by the `enr` package.  As currently stands, this PR implements a `defaultCrypto` object using `noble/hashes` and `noble/secp256k1` that are pure javascript implementation and are somewhat slower than the `bcrypto` equivalents but can easily be replaced with WASM compiled alternatives such as `as-sha256` or `polkadot/wasm-crypto`.

Sidequest 1: Replace `bcrypto` AES cipher functionality with nodejs builtins that are overall 5-10% faster.
Sidequest 2: Replace `bcrypto` `randomBytes` function with `noble` equivalent


acolytec3:
@wemeetagain Here's the initial draft.  I haven't implemented the `wasm-crypto` alternative to `noble` yet (that is much more performant), but this basically what I'm thinking.

acolytec3:
Note, the CI is failing because of a quirk in `noble/hashes` related to [Node 18 and older versions](https://github.com/paulmillr/noble-secp256k1?tab=readme-ov-file#usage).   Would you be open to just bumping the nodejs version to 20 in CI to eliminate this issue?  If not, I'll do the required shim but seems kind of silly given node 18 is EOL at this point.

packages/discv5/src/enr/setV4Crypto.ts
```diff
@@ -1,6 +1,7 @@
...
+// This module has the side effect of setting the ENR crypto implementations to use the defaultCrypto
+// implementation at present.
+// TODO: Adjust this if we use a different implementation
 
-import { setV4Crypto } from "@chainsafe/enr";
-import * as bcryptoV4Crypto from "./bcryptoV4Crypto.js";
+import { setV4Crypto, defaultCrypto } from "@chainsafe/enr";
 
-setV4Crypto(bcryptoV4Crypto);
+setV4Crypto(defaultCrypto);
```
wemeetagain, line 7: You can get rid of this file since we no longer need to switch implementations (and remove reference in src/index.tx)

acolytec3: Do you want me to implement a wasm-crypto version of this so we have access to it in discv5?  We use [@polkadot/wasm-crypto](https://github.com/polkadot-js/wasm/tree/master/packages/wasm-crypto) in the ethereumjs-monorepo as a drop-in replacement for JS native cryptography in our client and it's something like 10x faster on average.  I haven't benchmarked it against bcrypto but it's maintained and has nice typescript definitions packaged with it.

acolytec3: Okay, maybe not 10x, but here's two key functions with benchmarks by vitest for `bcrypto`, `wasm-crypto`, and `noble` 
image

I guess `bcrypto` is still the fastest since it's native c code

acolytec3: I chatted with @scottypoi and we decided against adding in any extra cryptography libraries and just going with `noble`.  The main place we get a win is switching the AES stuff to nodejs builtins and then dropping `bcrypto` entirely.  

So, have gone ahead and removed the file above as suggested.

acolytec3:
> Note, the CI is failing because of a quirk in `noble/hashes` related to [Node 18 and older versions](https://github.com/paulmillr/noble-secp256k1?tab=readme-ov-file#usage). Would you be open to just bumping the nodejs version to 20 in CI to eliminate this issue? If not, I'll do the required shim but seems kind of silly given node 18 is EOL at this point.

@wemeetagain would you be open to bumping CI to run on node 20 or do you want me to apply the shim that `noble` says is required to run on node 18?  Note, ultralight is running on node 20 and above only so not an issue for us.

acolytec3:
Other than the CI nodejs environment question, this should be ready for review.

packages/discv5/test/unit/session/crypto.test.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 import { expect } from "chai";
-import secp256k1 from "bcrypto/lib/secp256k1.js";
-import { randomBytes } from "bcrypto/lib/random.js";
+import * as secp256k1 from "@noble/secp256k1";
```
wemeetagain, line 3: Small thing, I think this file should be using getDiscv5Crypto instead of noble directly.

wemeetagain (approved):
LGTM
Thanks for sticking with this over the years :)

PR #306, opened 2025-01-13
feat!: Convert `buffer` usage to `Uint8Array`

### LINKED ISSUE #303: `Buffer` usage
Once #302 gets merged, we remove `bcrypto` which I believe was the main reason for continuing to use `Buffer` instead of `Uint8Array` everywhere.  Is that correct or are there other reasons to continue using `Buffer` in this library?  If not, would you be open to a PR to remove all the `Buffer` to/from which is just an unnecessary set of conversions that muck up the APIs.

acolytec3:
Fixes #303 

Problems:
1) The `enr` packages references `globalThis.Buffer` in the base64 conversion utilities and this causes segmentation faults in the `Ultralight` hive tests
2) `Buffer` was only used for bytes (instead of `Uint8Array`) because of the requirements of `bcrypto` which only accepted `Buffer` as input.
3) The `bigint-buffer` package causes CI failures in Ultralight with arbitrary `napi-ok === false` errors that are indecipherable 

Solution:
- Remove all `Buffer` types and utilities and replace with equivalent Uint8Arrays
- Switch `toHex`/`fromHex` usage to `hexToBytes`/`bytesToHex` from `ethereum-cryptography`
- Switch all `utf8` conversion to `ethereum-cryptography` `utf8ToBytes`/`bytesToUtf8` utilities
- Remove all references to `globalThis.buffer` 
- Update base64 utilities conversion to use `scure/base` (author same as `@noble/hashes`)
- Remove `bigint-buffer` and replace with internal `bigIntToBytes`/`bytesToBigInt` conversion utilities

### DIFF
--- a/packages/discv5/src/util/toBuffer.ts
+++ b/packages/discv5/src/util/toBuffer.ts
@@ -1,4 +1,4 @@
-export function numberToBuffer(value: number, length: number): Buffer {
+export function numberToBuffer(value: number, length: number): Uint8Array {
   const res = Buffer.alloc(length);
   res.writeUIntBE(value, 0, length);
   return res;
--- a/packages/discv5/src/util/toBuffer.ts
+++ b/packages/discv5/src/util/toBuffer.ts
@@ -1,8 +1,3 @@
-export function toBuffer(arr: Uint8Array): Buffer {
-  if (arr instanceof Buffer) return arr;
-  return Buffer.from(arr.buffer, arr.byteOffset, arr.length);
-}
-
 export function numberToBuffer(value: number, length: number): Buffer {
   const res = Buffer.alloc(length);
   res.writeUIntBE(value, 0, length);
--- a/packages/discv5/src/util/toBuffer.ts
+++ b/packages/discv5/src/util/toBuffer.ts
@@ -1,9 +1,0 @@
-export function numberToBuffer(value: number, length: number): Uint8Array {
-  const res = Buffer.alloc(length);
-  res.writeUIntBE(value, 0, length);
-  return res;
-}
-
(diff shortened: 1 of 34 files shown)


acolytec3:
@wemeetagain Hopefully this isn't too far beyond what you had in mind with regard to #303.  Removing buffers provided an opportunity to fix 2 other pain points for Ultralight.  If the `bigint-buffer` performance drop you'll see from going over to the the bigint -> hex -> butes route is a concern, I'd be happy to look for ways to improve the performance though I think @paulm has done a pretty thorough job of squeezing all the performance he can out of that conversion process.

acolytec3:
@wemeetagain bump

packages/discv5/src/util/toBuffer.ts
```diff
@@ -1,9 +1,4 @@
-export function toBuffer(arr: Uint8Array): Buffer {
-  if (arr instanceof Buffer) return arr;
-  return Buffer.from(arr.buffer, arr.byteOffset, arr.length);
-}
-
-export function numberToBuffer(value: number, length: number): Buffer {
+export function numberToBuffer(value: number, length: number): Uint8Array {
```
wemeetagain, line 1: Is this function still used? maybe delete it or refactor the usage of Buffer here

wemeetagain (commented):
Generally LGTM, thanks for further simplifying the codebase.

Just had one small comment

packages/discv5/src/util/toBuffer.ts
acolytec3: Good catch.  Not sure how I missed that but will do.

acolytec3:
@wemeetagain should be ready for final review now.

wemeetagain (approved):

PR #308, opened 2025-04-04
fix:  BigInt conversion bug

ScottyPoi:
This is a fix for a bug in the `ENR` package that was causing discv5 errors in ultralight. 

`enr/src/util.ts` exports a helper function `bytesToBigInt(bytes: Uint8Array): bigint`

This function uses `bytesToHex` from `ethereum-cryptography/utils` to convert the `Uint8Array` to a hex string, and then `BigInt()` to convert the hex string to `bigint`.

The bug occurs when the length of `bytes` is 0.  `bytesToHex` returns an empty string, and trying to call `BigInt('0x')` throws an error.

Fix: simply return `0n` if `bytes.length === 0`

### DIFF
--- a/packages/enr/src/util.ts
+++ b/packages/enr/src/util.ts
@@ -32,5 +32,8 @@ export function bigintToBytes(n: bigint): Uint8Array {
 }
 
 export function bytesToBigint(bytes: Uint8Array): bigint {
+  if (bytes.length === 0) {
+    return 0n;
+  }
   return BigInt(`0x${bytesToHex(bytes)}`);
 }
--- a/packages/enr/src/util.ts
+++ b/packages/enr/src/util.ts
@@ -33,7 +33,7 @@ export function bigintToBytes(n: bigint): Uint8Array {
 
 export function bytesToBigint(bytes: Uint8Array): bigint {
   if (bytes.length === 0) {
-    return 0n;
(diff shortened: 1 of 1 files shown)


acolytec3:
@wemeetagain ^^ This was a bug introduced by my `buffer` PR a while back. @fjl found it with the new discv5 hive tests he's working on

ScottyPoi:
> @wemeetagain ^^ This was a bug introduced by my `buffer` PR a while back. @fjl pointed out we had an issue with the new discv5 hive tests he's working on and @ScottyPoi found the fix

this doesn't actually fix any of those hive tests.  but it did fix that particular error :smile:

packages/enr/src/util.ts
```diff
@@ -32,5 +32,8 @@ export function bigintToBytes(n: bigint): Uint8Array {
 }
 
 export function bytesToBigint(bytes: Uint8Array): bigint {
+  if (bytes.length === 0) {
+    return 0n;
```
wemeetagain, line 36: Small thing, may be better to rather do BigInt(0) which has a nicer failure mode for incompatible browsers/envs

ScottyPoi: :heavy_check_mark:

wemeetagain (approved):

Issue #309, opened 2025-04-07
Hive tests fail due to RequestId conversion


ScottyPoi:
Issue causing `hive` test failures in `discv5` suite:
View the test [here](https://portal-hive-experimental.ethdevops.io/suite.html?suiteid=1744053282-6aec00015b9169ec1cd558c342a4fe35.json&suitename=discv5#test-6)

Request sent with `id: 00000001`
Expects response with `id: 00000001`
Actual response has `id: 01`

----

This issue above is created due to conversion from `bytes` to `bigint` back to `bytes`

When a request is decoded, the `RequestId` gets converted from `bytes` to a `bigint`  via `bytesToBigint`
Then when the response is encoded, that `RequestId`  (stored as a `bigint`)  is converted back to `bytes`  via `bigintToBytes`

`bytesToBigint(Uint8Array([0, 0, 0, 1]) = 1n`
`bigintToBytes(1n) = Uint8Array([1])`

ScottyPoi:
Fixed by #312

PR #312, opened 2025-04-10
fix: leading zeros stripped from request id

ScottyPoi:
Fix for issue #309 

This is a fix for a number of hive tests.  Here is an [example](https://portal-hive-experimental.ethdevops.io/suite.html?suiteid=1744230445-096c31f89428ef5aa5e2c28dc5b7cffe.json&suitename=discv5#test-6)

The tests fail with errors like this:
` wrong request ID 01 in TALKRESP, want 00000001`

----

This is happening because when we `decode` an incoming `TalkRequest`, we store the `request.id` in memory as a `bigint`, using a `bytesToBigint` conversion.  

Then when we `encode` the `TalkResponse`, we convert the `request.id` back from a `bigint` to bytes using `bigintToBytes`.  

The requesting node compares the `response.id` to the `request.id` that it sent, and since the bytes are not equal, it throws a `wrong request ID` error.

In the example above:

`request.id = Uint8Array [0, 0, 0, 0, 0, 1]`
`bytesToBigint(request.id) = 1n`
`response.id = bigintToBytes(1n) = Uint8Array [1]`

----

This fix proposed here: 

- Changes the type of `RequestId` to `Uint8Array`
- Removes the `bytes<>bigint` conversions from `encode` and `decode`.  
- `createRequestId` now returns a `Uint8Array`.  
- Changes the comparisons of request.id to response.id to `equalsBytes` instead of `===` and `!==`.  
- Still use the `bigint` conversion as request map keys since Uint8Array cannot be a map key

### DIFF
--- a/packages/discv5/src/message/create.ts
+++ b/packages/discv5/src/message/create.ts
@@ -1,5 +1,5 @@
 import { randomBytes, toBytes } from "@noble/hashes/utils";
-import { bytesToBigint, SequenceNumber, ENR } from "@chainsafe/enr";
+import { SequenceNumber, ENR } from "@chainsafe/enr";
 
 import {
   RequestId,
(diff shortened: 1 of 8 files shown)


wemeetagain (approved):
LGTM

PR #313, opened 2025-04-22
fix: should not respond to oversized id

ScottyPoi:
Fix for hive test in `devp2p/discv5` suite: `PingLargeRequestID`

Test fails because node responds despite oversized `requestId`
```
 This test sends PING with a 9-byte request ID, which isn't allowed by the spec.
 The remote node should not respond.
 (370c16d0a4762252) >> PING/v5
 (370c16d0a4762252) << WHOAREYOU/v5
 (370c16d0a4762252) >> PING/v5
 response with oversized request ID
```

Solution: use custom function `decodeRequestId` during `decode` functions.  `decodeRequestId` performs bytelength check and throws an error if length > 8

### DIFF
--- a/packages/discv5/src/message/decode.ts
+++ b/packages/discv5/src/message/decode.ts
@@ -60,7 +60,7 @@ function decodePing(data: Uint8Array): IPingMessage {
   }
   return {
     type: MessageType.PING,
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
Lgtm

PR #315, opened 2025-06-14
fix: better type safety and no null checks

ChefBingbong:


wemeetagain:
Hm, type-safety is a pretty objective measurement. I don't see any improvement from this PR. Reason being that optional chaining (a builtin language feature) has a quite clear meaning, and it is well modeled in TypeScript. Your change (1) is roughly  (but not entirely!) equivalent, so it is mostly a lateral move wrt type-safety.

Wrt code cleanliness / readability / maintainability, this is somewhat subjective. And here I disagree with the changes you have. With 1, this is simply more verbose with no clear benefit. We think optional chaining is a useful syntax sugar and should be used in place of nullish checks in `if` or ternarys. We're firm on this. With 2, this is somewhat more subjective. But the intent in code here is to be explicit. We don't need to generically collect and iterate over sockets/configs, it doesn't even save us any lines of code! Because of the really limited/subjective benefit, I choose not to accept this change either.

PR #339, opened 2026-08-09
fix(enr): handle short port values without throwing

nflaig:
## What changed

- normalize ENR port bytes before `getProtocolValue()` reads them
- accept valid one-byte and two-byte big-endian port values
- return `undefined` for empty, zero, or oversized port values instead of throwing
- keep low ports produced by the existing setters usable by `getLocationMultiaddr()`
- add ENR-level regression tests and direct coverage for discv5's `getSocketAddressOnENR()` path

## Why

A mainnet Lodestar v1.45.0 node using `@chainsafe/enr@6.0.1` and `@chainsafe/discv5@12.0.1` received a signed remote ENR with a short `udp` value. During session establishment, `getSocketAddressOnENR()` accessed `enr.udp`, and `getProtocolValue()` threw:

```
uncaughtException: Encoded protocol length should be 2
at getProtocolValue
at get udp
at getSocketAddressOnENRByFamily
at getSocketAddressOnENR
at SessionService.onEstablished
```

ENR ports are big-endian integers, so one-byte values are valid. Invalid values received from untrusted peers should make the ENR non-contactable rather than escape as an exception.

This follows up on the normalization added in #333 and applies it to the direct port getters used by discv5.

Related: ChainSafe/lodestar#7445, ChainSafe/lodestar#9310.

## Impact

Valid compact ports now decode correctly. Empty, zero, and oversized values are ignored without disrupting the discv5 session service. Normal two-byte ports are unchanged.

## Validation

- `pnpm build`
- `pnpm lint` (passes; two existing warnings outside this change)
- `pnpm check-types`
- `pnpm test:unit` — 28 ENR tests and 65 discv5 tests pass
- `pnpm test:e2e` — 5 tests pass

### DIFF
--- a/packages/enr/src/enr.ts
+++ b/packages/enr/src/enr.ts
@@ -181,18 +181,10 @@ export function getIPValue(
 }
 
 export function getProtocolValue(kvs: ReadonlyMap<ENRKey, ENRValue>, key: string): number | undefined {
-  const raw = normalizePortBytes(kvs.get(key));
-  if (raw) {
-    return (raw[0] << 8) + raw[1];
-  }
-  return undefined;
-}
-
-function normalizePortBytes(raw: Uint8Array | undefined): Uint8Array | undefined {
+  const raw = kvs.get(key);
   if (!raw || raw.length === 0 || raw.length > 2) return undefined;
-  if (raw.length === 1) return raw[0] === 0 ? undefined : new Uint8Array([0, raw[0]]);
-  if (raw[0] === 0 && raw[1] === 0) return undefined;
-  return raw;
+  const port = raw.length === 1 ? raw[0] : (raw[0] << 8) + raw[1];
+  return port === 0 ? undefined : port;
 }
 
 export function portToBuf(port: number): Uint8Array {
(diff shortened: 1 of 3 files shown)


packages/enr/src/enr.ts
```diff
@@ -181,20 +181,17 @@ export function getIPValue(
 }
 
 export function getProtocolValue(kvs: ReadonlyMap<ENRKey, ENRValue>, key: string): number | undefined {
-  const raw = kvs.get(key);
+  const raw = normalizePortBytes(kvs.get(key));
```
wemeetagain: maybe this should return the number (eg do the bitshift + add)

packages/enr/src/enr.ts
```diff
@@ -182,20 +182,9 @@ export function getIPValue(
 
 export function getProtocolValue(kvs: ReadonlyMap<ENRKey, ENRValue>, key: string): number | undefined {
```
nflaig: should this be renamed to `getPortValue`? or otherwise could rename the `port` variable below if we think this will return anything other than a port in the future

packages/enr/src/enr.ts
nflaig: updated in https://github.com/ChainSafe/discv5/pull/339/changes/4892067ec502938eba4eb61ed5f3159e3d2028b7

packages/enr/src/enr.ts
wemeetagain: yes, getPortValue is better

nflaig: renamed in https://github.com/ChainSafe/discv5/pull/339/changes/fc93724f8b6ad0fc9c6764160538cc0dae04dc60, kept alias in case someone is using that function

wemeetagain (approved):

